PR #4, opened 2020-11-12
Fix calculation of range for candidates near EOF

radeksimko:
Fixes wrong range calculation for newText/snippet when the position is near EOF, as reported in https://github.com/hashicorp/terraform-ls/issues/295

### DIFF
--- a/decoder/candidates.go
+++ b/decoder/candidates.go
@@ -177,6 +177,15 @@ func nameTokenRangeAtPos(tokens hclsyntax.Tokens, pos hcl.Pos) (hcl.Range, error
 			}
 			return hcl.Range{}, fmt.Errorf("token is %s, not Ident", t.Type.String())
 		}
+
+		// EOF token has zero length
+		// so we just compare start/end position
+		if t.Type == hclsyntax.TokenEOF && t.Range.Start == pos && t.Range.End == pos && i > 0 {
+			previousToken := tokens[i-1]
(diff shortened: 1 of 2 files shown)


decoder/candidates.go
```diff
@@ -177,6 +177,13 @@ func nameTokenRangeAtPos(tokens hclsyntax.Tokens, pos hcl.Pos) (hcl.Range, error
 			}
 			return hcl.Range{}, fmt.Errorf("token is %s, not Ident", t.Type.String())
 		}
+		// EOF has zero length
```
appilon, line 180: What does this mean, EOF has zero length?

appilon (approved):

decoder/candidates.go
radeksimko: I can tune the comment, it means the length of the _EOF token_ is zero, because it's not really anything visible in the editor.

This causes `ContainsPos()` to always return `false` since it's looking in a zero-length range.

appilon: How can an EOF _token_ have zero length? Wouldn't the existence of a token, have at least length 1? 🤔

radeksimko: This is not byte length, it's the length of the HCL range, so basically a _(visible) character_ slice length. EOF is technically invisible, similar to BOM and other special bytes. You wouldn't see those rendered in a typical editor.

PR #6, opened 2020-12-07
decoder: Introduce SemanticTokensInFile

radeksimko:
It is possible that we'll need to change the interface in the future to account for the other methods defined in the protocol where tokens are calculated for a particular range, or some kind of "delta" is calculated.

I would however prefer to leave that concern for later and still enable shipping of an MVP.
We still treat this library as "not stable", i.e. we can still change interfaces in a non-compatible way.


decoder/semantic_tokens.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+
+		tokens = append(tokens, lang.SemanticToken{
+			Type:      lang.TokenBlockType,
+			Modifiers: modifiers,
+			Range:     block.TypeRange,
+		})
+
+		for i, labelRange := range block.LabelRanges {
+			if i+1 > len(blockSchema.Labels) {
```
appilon, line 92: Feels odd to me, but I'm not 100% up to speed with what is going on

appilon (approved):
overall lgtm

decoder/semantic_tokens.go
radeksimko: Good catch, there should be `>=`. This is just to catch a situation of e.g. 3rd label declared for `resource` which has schema for just 2 labels.

radeksimko: Actually this works just fine as is. This is what happens in that situation:

```go
if i+1 > len(blockSchema.Labels)
```
```go
// 3rd label has index of 2
// schema only declares 2 labels
if 2+1 > 2
```
```go
if 3 > 2 # true
```

PR #7, opened 2021-01-28
Introduce method for querying symbols across files

radeksimko:
This provides support for `workspace/symbol` method on the language server side:

https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_symbol

## Query Matching

Currently we'd do a simple exact match on the symbol's "name" - e.g. `resource "aws_vpc" "main"`, but I reckon longer term we may want to add some loose matching too, or perhaps compare the query against the common `block.label1.label2` pattern.

## Query Scope

Currently we only query the top-level symbols, but we may consider querying nested symbols, i.e. nested attributes and blocks too eventually. It just isn't clear to me yet what the UX implications of that would be in bigger modules with many blocks.

### DIFF
--- a/decoder/symbols.go
+++ b/decoder/symbols.go
@@ -89,5 +89,9 @@ func symbolsForBody(body *hclsyntax.Body) []Symbol {
 		})
 	}
 
+	sort.SliceStable(symbols, func(i, j int) bool {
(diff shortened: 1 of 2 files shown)


aeschright (approved):
Looks like an appropriate scope to start with 👍

PR #8, opened 2021-01-28
Introduce LinksInFile method

radeksimko:
This is in preparation for implementation of `textDocument/documentLink` on the language server side:
https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_documentLink

https://github.com/hashicorp/terraform-schema/pull/10 adds the actual links to the Registry for Terraform providers and depends on this PR

Initially I decided to only implement this for _dependent_ block bodies, which in case of Terraform applies to provider, resource and data source blocks as that's the part of configuration which has more easily linkable and versioned docs.

Relatedly I also limited the implementation initially for the 1st level, as we don't have dependent bodies at any deeper levels.

### DIFF
--- a/decoder/decoder.go
+++ b/decoder/decoder.go
@@ -20,6 +20,14 @@ type Decoder struct {
 	rootSchema    *schema.BodySchema
 	rootSchemaMu  *sync.RWMutex
 	maxCandidates uint
+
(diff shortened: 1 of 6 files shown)


aeschright (approved):
👍

PR #9, opened 2021-02-16
ci: Enable testing in GitHub Actions

radeksimko:
This enables continuous testing in GitHub Actions and also effectively enables CodeCov (alongside https://github.com/hashicorp/hcl-lang/pull/10) by uploading the coverage file to the relevant codecov API.


.gitattributes
```diff
@@ -0,0 +1 @@
+*.go text eol=lf
```
radeksimko, line 1: This is to avoid EOL diffs when the repo is cloned on Windows.

aeschright (approved):

PR #21, opened 2021-03-01
refactor: lang.Reference -> lang.Address

radeksimko:
"address" better reflects the contents of that slice and how it is used.

This is in preparation for supporting traversal expressions where reference can also have a different meaning.


aeschright (approved):
👍

PR #23, opened 2021-03-10
Add support for ObjectExpr expression type

radeksimko:
This is to support recent changes in core schema introduced in Terraform 0.15 where an object can contain traversals - i.e. not literals.

### DIFF
--- a/decoder/expression_candidates.go
+++ b/decoder/expression_candidates.go
@@ -56,6 +56,37 @@ func constraintsAtPos(expr hcl.Expression, constraints ExprConstraints, pos hcl.
 				Filename: eType.Range().Filename,
 			}
 		}
+	case *hclsyntax.ObjectConsExpr:
+		oe, ok := constraints.ObjectExpr()
+		if ok {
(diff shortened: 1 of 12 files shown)


decoder/expression_candidates.go
```diff
@@ -108,9 +140,10 @@ func constraintToCandidates(constraint schema.ExprConstraint, editRng hcl.Range)
 		})
 	case schema.MapExpr:
 		candidates = append(candidates, lang.Candidate{
-			Label:  fmt.Sprintf(`{ key =%s}`, labelForConstraints(c.Elem)),
-			Detail: c.FriendlyName(),
-			Kind:   lang.MapCandidateKind,
+			Label:       fmt.Sprintf(`{ key =%s}`, labelForConstraints(c.Elem)),
+			Detail:      c.FriendlyName(),
+			Description: c.Description,
```
radeksimko, line 145: Small (unrelated) fix I made as I was reading through the code, probably doesn't affect much in practice though as we don't have too many map expressions with description.

aeschright (approved):

PR #29, opened 2021-03-26
fix: avoid crashes on unknown keys

radeksimko:
This addresses https://github.com/hashicorp/terraform-ls/issues/447


decoder/symbols.go
```diff
@@ -125,7 +125,7 @@ func nestedSymbolsForExpr(expr hcl.Expression) []Symbol {
 	case *hclsyntax.ObjectConsExpr:
 		for _, item := range e.Items {
 			key, _ := item.KeyExpr.Value(nil)
-			if key.IsNull() || key.Type() != cty.String {
+			if key.IsNull() || !key.IsWhollyKnown() || key.Type() != cty.String {
```
radeksimko, line 128: This is the actual fix, other code is just tests and refactoring for consistency.

aeschright (approved):
👍

PR #32, opened 2021-04-14
Avoid completing duplicate label candidates

radeksimko:
https://github.com/hashicorp/hcl-lang/pull/25 apparently uncovered another edge case in completion of labels.

Specifically with #25 applied we'd "correctly" base our suggestions for label on all available dependency keys, which are essentially duplicated, as is demonstrated in the attached test.

I am not sure yet what's the best way of approaching this (admittedly not common) situation, but I do think that user most likely wouldn't expect duplicated entries, especially when they have the exact same meaning.

Screenshot 2021-04-14 at 20 30 30

### DIFF
--- a/decoder/candidates_test.go
+++ b/decoder/candidates_test.go
@@ -643,6 +643,98 @@ func TestDecoder_CandidatesAtPos_emptyLabel(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


aeschright (approved):
👍

PR #40, opened 2021-05-17
Implement Copy methods on schema structs & lang.References

radeksimko:
### Why

Certain structs need to be copied regularly within the language server.

Currently this is done via `copystruct` which is easy (from implementation/readability perspective) and it works, but it also comes at a cost of higher CPU usage. Copying a few structs every once in while this way would probably work just fine (from CPU usage perspective), but in LS this operation is pretty common operation and so such small detail causing small difference does add up.

This patch therefore aims to help reduce CPU consumption downstream by avoiding reflection and help address https://github.com/hashicorp/terraform-ls/issues/509

--- 

### Implementation Notes

I briefly considered making shallow copies within all `Copy()` functions and then overwriting the "problematic" fields, such as maps or slices - which would be better from maintenance & readability (no need to list out every field).

For example

```go
func (lv LiteralValue) Copy() ExprConstraint {
	newLv := lv
	return newLv
}
```

However this would also mean, that we could (unintentionally) end up making a shallow copy a new map/slice/pointer.

The current implementation therefore won me over due to side-effects being "safer" or more obvious. i.e. if someone in the future forgets to add a field I would prefer to it _not_ being copied at all rather than silently making a shallow copy.


aeschright (approved):
👍

PR #42, opened 2021-06-01
Add `IsSensitive` to `AttributeSchema`

beandrad:
So that sensitive attributes can be flagged.

### DIFF
--- a/decoder/attribute_candidates.go
+++ b/decoder/attribute_candidates.go
@@ -28,23 +28,24 @@ func attributeSchemaToCandidate(name string, attr *schema.AttributeSchema, rng h
 }
 
 func detailForAttribute(attr *schema.AttributeSchema) string {
-	var detail string
+	details := []string{}
+
 	if attr.IsRequired {
-		detail = "Required"
+		details = append(details, "required")
 	} else if attr.IsOptional {
-		detail = "Optional"
+		details = append(details, "optional")
+	}
+
+	if attr.IsSensitive {
+		details = append(details, "sensitive")
 	}
 
 	friendlyName := attr.Expr.FriendlyName()
 	if friendlyName != "" {
-		if detail != "" {
-			detail = strings.Join([]string{detail, friendlyName}, ", ")
-		} else {
-			detail = friendlyName
-		}
+		details = append(details, friendlyName)
 	}
 
-	return detail
+	return strings.Join(details[:], ", ")
 }
 
 func snippetForAttribute(name string, attr *schema.AttributeSchema) string {
--- a/decoder/body_decoder_test.go
+++ b/decoder/body_decoder_test.go
@@ -127,7 +127,7 @@ func TestDecoder_CandidateAtPos_computedAttributes(t *testing.T) {
 		List: []lang.Candidate{
 			{
 				Label:  "attr2",
-				Detail: "Optional, number",
+				Detail: "optional, number",
 				TextEdit: lang.TextEdit{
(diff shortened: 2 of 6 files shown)


decoder/attribute_candidates.go
```diff
@@ -35,6 +35,10 @@ func detailForAttribute(attr *schema.AttributeSchema) string {
 		detail = "Optional"
 	}
 
+	if attr.IsSensitive {
```
beandrad, line 38: not sure about this...

radeksimko (commented):
Thanks for the PR!

We'll need to reflect this new field also in hover data: https://github.com/hashicorp/hcl-lang/blob/b62f3a9965765599ec3bb707d628467b6afdd4a7/decoder/attribute_candidates.go#L30

Also would you mind adding some test cases for the public `CandidatesAtPos` here
https://github.com/hashicorp/hcl-lang/blob/b62f3a9965765599ec3bb707d628467b6afdd4a7/decoder/candidates_test.go#L803-L807

and similar for `HoverAtPos` here
https://github.com/hashicorp/hcl-lang/blob/b62f3a9965765599ec3bb707d628467b6afdd4a7/decoder/hover_test.go#L328-L332

One case in each (with sensitive attribute) should suffice; if you want to be thorough you could add one with `AttributeSchema` where `Expr` is `ObjectExpr` with sensitive attribute nested within, although I'm not sure what's the current handling of it as nested expressions aren't well supported yet - but it's still nice to have a test that reflects the current behaviour - and will fail (and be fixed) as part of implementing https://github.com/hashicorp/terraform-ls/issues/496

decoder/attribute_candidates.go
radeksimko: Also for variable schema specifically we don't actually know (yet) whether a variable is required or optional without parsing the `default` - which we decided not to parse initially, so we should most certainly account for this eventuality here.

radeksimko:
As per conversation in https://github.com/hashicorp/terraform-schema/pull/49#issuecomment-852166335 we should also add some tests which test `AttributeSchema` with no explicit required/optional during completion or hover. I think the validation should keep reporting it as failure/error, but we can treat it as "soft enforcement" for pragmatic reasons, so the decoder should account for this edge case and test for it.

decoder/attribute_candidates.go
radeksimko: You could probably create a string slice, such as `details []string` and then `strings.Join()` everything together upon return 🤔  and I wouldn't worry too much about the capital letter in `sensitive` - it's an edge case anyway and it's kinda debatable whether optional/required should be capitalized anyway in this context 🤷🏻

decoder/attribute_candidates.go
```diff
@@ -28,23 +28,24 @@ func attributeSchemaToCandidate(name string, attr *schema.AttributeSchema, rng h
 }
 
 func detailForAttribute(attr *schema.AttributeSchema) string {
-	var detail string
+	details := []string{}
+
 	if attr.IsRequired {
-		detail = "Required"
+		details = append(details, "required")
```
radeksimko, line 34: This is a reasonable change, but just FYI there will be some existing test data in the need of update potentially in `terraform-schema` and in `terraform-ls` 😅

radeksimko (approved):

decoder/attribute_candidates.go
beandrad: sounds exciting  🙈

PR #47, opened 2021-06-07
Add type declarations for variable `type`

beandrad:
So that completion, hover and semantic-tokens work for variable `type`.

Related to hashicorp/terraform-ls#490

### DIFF
--- a/decoder/semantic_tokens.go
+++ b/decoder/semantic_tokens.go
@@ -218,6 +218,28 @@ func (d *Decoder) tokensForExpression(expr hclsyntax.Expression, constraints Exp
 
 			}
 		}
+		_, ok = constraints.TypeDeclarationExpr()
+		if ok {
+			tokens = append(tokens, lang.SemanticToken{
+				Type:      lang.TokenTypePrimitive,
+				Modifiers: []lang.SemanticTokenModifier{},
+				Range:     expr.Range(),
+			})
+		}
+	case *hclsyntax.FunctionCallExpr:
+		_, ok := constraints.TypeDeclarationExpr()
+		if ok {
+			tokens = append(tokens, lang.SemanticToken{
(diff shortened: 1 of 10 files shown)


radeksimko (commented):
This is a good start.

Aside from my in-line comments, I think this new expression type will also need to be reflected in `SemanticTokensInFile` so that we can highlight the types semantically. Relatedly we'll also need to introduce new token types for this purpose, as mentioned in https://github.com/hashicorp/terraform-ls/issues/490

https://github.com/hashicorp/hcl-lang/blob/f2d6b65f96769bbb0ee5c1566298efab56001f20/decoder/semantic_tokens.go#L129

decoder/semantic_tokens.go
```diff
@@ -218,6 +218,28 @@ func (d *Decoder) tokensForExpression(expr hclsyntax.Expression, constraints Exp
...
+				Range:     expr.Range(),
+			})
+		}
+	case *hclsyntax.FunctionCallExpr:
+		_, ok := constraints.TypeDeclarationExpr()
+		if ok {
+			tokens = append(tokens, lang.SemanticToken{
+				Type:      lang.TokenTypeCapsule,
+				Modifiers: []lang.SemanticTokenModifier{},
+				Range:     expr.Range(),
```
radeksimko, line 235:
```suggestion
				Range:     eType.NameRange,
```
Here I think that on the client/editor side we probably wouldn't want to highlight brackets with the same colour as the type name itself. Usually brackets, dots, commas etc. are not reported as highlightable tokens and they're already covered by the static grammar on client side.

Also the LSP isn't super clear about overlapping tokens and VS Code AFAIK doesn't handle those well.

radeksimko (commented):
LGTM aside from that one minor in-line comment.

Thank you for making those changes.

radeksimko (approved):

PR #52, opened 2021-06-11
Add `optional` in hover content for object attrs

beandrad:
To display whether an object attribute is optional; for instance, in the
object variable below, `opt_attr` is optional.

```
variable "test" {
  type = object({
      opt_attr = optional(string)
    }
  )
}
```


radeksimko (approved):
Looks good, thanks for the PR! 👍🏻

PR #58, opened 2021-06-24
Introduce `ReferenceOriginAtPos()`

radeksimko:
This is to provide support for "go to definition" and "go to declaration", as described in https://github.com/hashicorp/vscode-terraform/issues/671

TODO:

 - tests
 - (separate PR) collect missing references to list and map block types


radeksimko:
Replaced by #62

PR #89, opened 2021-09-28
Prefill Required Attributes on Block Completion

jpogran:


decoder/block_candidates.go
```diff
@@ -54,18 +54,48 @@ func detailForBlock(block *schema.BlockSchema) string {
...
+			if l.IsDepKey {
+				labels += fmt.Sprintf(` "${%d}"`, placeholder)
+			} else {
+				labels += fmt.Sprintf(` "${%d:%s}"`, placeholder, l.Name)
+			}
+			placeholder++
+		}
+
+		return fmt.Sprintf("%s%s {\n  ${%d}\n}", blockType, labels, placeholder)
+	}
```
radeksimko, line 100:
```suggestion
	}
	
	labels := ""
	placeholder := 1

	for _, l := range block.Labels {
		if l.IsDepKey {
			labels += fmt.Sprintf(` "${%d}"`, placeholder)
		} else {
			labels += fmt.Sprintf(` "${%d:%s}"`, placeholder, l.Name)
		}
		placeholder++
	}

	return fmt.Sprintf("%s%s {\n  ${%d}\n}", blockType, labels, placeholder)
```
Nitpick: I think the else block here is redundant since you return early from the other branch.

decoder/candidates_test.go
```diff
@@ -421,7 +421,7 @@ func TestDecoder_CandidatesAtPos_rightHandSide(t *testing.T) {
 		},
 	}
 	testConfig := []byte(`myblock "foo" {
-  num_attr = 
+  num_attr =
```
radeksimko, line 424: Was this intentional?

decoder/label_candidates.go
```diff
@@ -81,6 +95,19 @@ func (d *Decoder) labelCandidatesFromDependentSchema(idx int, db map[schema.Sche
 	return candidates, nil
 }
 
+func generateSnippet(label string, attr map[string]*schema.AttributeSchema) string {
```
radeksimko, line 98:
```suggestion
func generateRequiredFieldsSnippet(label string, attr map[string]*schema.AttributeSchema) string {
```
We have other functions which generate snippets and functions in Go are package-wide, so I think the name here should help differentiate this function from the others.

decoder/label_candidates.go
```diff
@@ -35,44 +37,56 @@ func (d *Decoder) labelCandidatesFromDependentSchema(idx int, db map[schema.Sche
...
+			// to know how to do completion within a block
+			// but this doesn't matter when completing the label itself
+			// unless/until we're also completing the dependent attributes.
+			if _, ok := foundCandidateNames[label.Value]; ok {
+				continue
+			}
+
+			te := lang.TextEdit{}
+			if d.PrefillRequiredAttributes {
+				snippet := generateSnippet(label.Value, bodySchema.Attributes)
```
radeksimko, line 65: As mentioned in Slack we'll also need to reflect "required" blocks here, i.e. range over `bodySchema.Blocks` and add any blocks with `MinItems > 0`.

One related question which may also come up is whether we also want to generate required attributes _within_ these blocks. Personally, I think that even empty blocks would be useful as a first iteration (i.e. I wouldn't worry about it _too much_) - but if you can get it all working then great!

btw. The story of blocks in provider SDK is historically tricky and they will likely be phased out (see https://github.com/hashicorp/terraform-plugin-framework/issues/85), but we should still support them, also more importantly `hcl-lang` doesn't focus only on provider schema, nor just Terraform.

decoder/label_schema_test.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/hashicorp/hcl-lang/schema"
+	"github.com/hashicorp/hcl/v2"
+	"github.com/hashicorp/hcl/v2/hclsyntax"
+	"github.com/zclconf/go-cty/cty"
+)
+
+func TestCandidatesAtPos_prefillRequiredBlocks(t *testing.T) {
+	tests := []struct {
+		name                      string
+		fileName                  string
```
radeksimko, line 17: nitpick: I think the filename doesn't matter in the context of testing this feature, so you could probably just hard-code it below instead of duplicating it in every test case.

decoder/decoder.go
```diff
@@ -28,6 +28,8 @@ type Decoder struct {
 	utmMedium string
 	// utm_content parameter, e.g. documentHover or documentLink
 	useUtmContent bool
+
+	PrefillRequiredAttributes bool
```
radeksimko, line 32:
```suggestion
	// PrefillRequiredFields enriches label-based completion candidates with required attributes and blocks
	PrefillRequiredFields bool
```
nitpick: "attributes -> fields" is a minor name change, but I think it better reflects that we're completing both attributes and blocks.

radeksimko (commented):
Aside from the completion of blocks I only have very minor stylistic suggestions. The PR overall looks pretty good! Well done.

decoder/candidates_test.go
jpogran: Nope. Do not have `"files.trimTrailingWhitespace": true,` and edit a file, then decide not to do something with that file 😂

decoder/label_schema_test.go
```diff
@@ -0,0 +1,256 @@
...
+								Byte:   13,
+							},
+						},
+						NewText: "aws_instance\" \"${2:type}\" {\n\tami = \"${3:value}\"\n\t${0}",
+						Snippet: "aws_instance\" \"${2:type}\" {\n\tami = \"${3:value}\"\n\t${0}",
+					},
+					Kind: lang.LabelCandidateKind,
+				},
+			}),
+		},
```
radeksimko, line 230: Do you mind adding one more test which exercises pre-filling of required attributes within a nested block, or modifying the existing one with `spec` block?

schema/body_schema.go
```diff
@@ -50,6 +51,24 @@ func NewBodySchema() *BodySchema {
 	}
 }
 
+func (as *BodySchema) SortAttributes() []string {
```
radeksimko, line 54:
```suggestion
func (as *BodySchema) AttributeNames() []string {
```
I guess `SortedAttributeNames` would also be a reasonable name, but either way I think that the name should communicate that we're just _extracting_ attribute names, not modifying any kind of internal order.

schema/body_schema.go
```diff
@@ -50,6 +51,24 @@ func NewBodySchema() *BodySchema {
...
+	keys := make([]string, 0, len(as.Attributes))
+	for k := range as.Attributes {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	return keys
+}
+
+func (as *BodySchema) SortBlocks() []string {
```
radeksimko, line 63:
```suggestion
func (as *BodySchema) BlockTypes() []string {
```
Similarly `SortedBlockTypes` would also be a reasonable name here, but we can also just imply that's sorted as most slices in Go should be anyway.

decoder/label_candidates.go
```diff
@@ -81,6 +98,46 @@ func (d *Decoder) labelCandidatesFromDependentSchema(idx int, db map[schema.Sche
...
+			placeholder++
+		}
+	}
+
+	blockNames := bodySchema.SortBlocks()
+	for _, blockName := range blockNames {
+		blockSchema := bodySchema.Blocks[blockName]
+		if blockSchema.MinItems > 0 {
+			snippetText += fmt.Sprintf("%s%s {\n", indent, blockName)
+			snippetText += generateRequiredFieldsSnippet("", blockSchema.Body, blockSchema.Labels, placeholder, indentCount+1)
```
radeksimko, line 133: Correct me if I'm wrong here but I have a sense that this might not work for blocks of map type due to how we assume that all inner blocks never have labels? Do you mind adding a test with a block of `Type: schema.BlockTypeMap`? 🤔 

Blocks of map type generally always have a single label and while it's not super common for these blocks to have `MinItems` it's not impossible.

decoder/label_candidates.go
```diff
@@ -2,22 +2,25 @@ package decoder
...
-func (d *Decoder) labelCandidatesFromDependentSchema(idx int, db map[schema.SchemaKey]*schema.BodySchema, prefixRng, editRng hcl.Range) (lang.Candidates, error) {
+func (d *Decoder) labelCandidatesFromDependentSchema(idx int, db map[schema.SchemaKey]*schema.BodySchema, prefixRng, editRng hcl.Range, block *hclsyntax.Block, labelSchemas []*schema.LabelSchema) (lang.Candidates, error) {
 	candidates := lang.NewCandidates()
 	candidates.IsComplete = true
 	count := 0
 
 	foundCandidateNames := make(map[string]bool, 0)
 
 	prefix, _ := d.bytesFromRange(prefixRng)
+	// newr, _ := d.bytesFromRange(hcl.RangeBetween(prefixRng, block.OpenBraceRange))
```
radeksimko, line 23: 👀

decoder/expression_candidates.go
```diff
@@ -673,6 +673,59 @@ func snippetForExprContraints(placeholder uint, ec schema.ExprConstraints) strin
...
+		}
+		return "[\n  ${0}\n]"
+	}
+	if t, ok := e.TupleExpr(); ok {
+		// TODO: multiple constraints?
+		ec := ExprConstraints(t.Elems[0])
+		if ec.HasKeywordsOnly() {
+			return "[ ${0} ]"
+		}
+		return "[\n  ${0}\n]"
```
radeksimko, line 715: Nitpick - also I'm aware that I may be still commenting on WIP here - so you are probably already aware (in which case ignore me! 😄  ) - but the `${0}` can be replaced with a type-aware snippet, as below.

decoder/block_candidates.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/hashicorp/hcl/v2"
 )
 
-func blockSchemaToCandidate(blockType string, block *schema.BlockSchema, rng hcl.Range) lang.Candidate {
+func blockSchemaToCandidate(blockType string, block *schema.BlockSchema, rng hcl.Range, prefillRequiredAttributes bool) lang.Candidate {
```
radeksimko, line 12: Nitpick: we might benefit from turning this function into a method on the `Decoder` as that would avoid the need for that extra argument.

schema/body_schema.go
jpogran: Apparently great minds think alike. You've already implemented these [here](https://github.com/hashicorp/hcl-lang/blob/4bf451bad31e52c5575261a27966b692f1c564de/decoder/body_candidates.go#L85-L105). Do you have a preference for that style vs this style? If we decide on an implementation, would it make sense to move the reference in `body_candidates` to use the interface method?

radeksimko: Ah, right, I completely forgot about these functions! I don't have too strong opinion, but my temptation would be to use the functions inside the `decoder` package.

The assumption/expectation I'm working with is that (from consumer perspective, such as `terraform-schema`) when declaring the schema it shouldn't matter how attributes are sorted, it's only when you decode them when it matters.

I also went as far as "redeclaring" the `ExpressionConstraint` schema type to reflect the same train of thoughts there: https://github.com/hashicorp/hcl-lang/blob/4bf451bad31e52c5575261a27966b692f1c564de/decoder/expression_constraints.go#L9 but I think that for sorting alone we shouldn't need it.

decoder/block_candidates.go
jpogran: Made it into a method on `Decoder`, but didn't move the method to `decoder.go` because otherwise there is nothing left in this file

decoder/expression_candidates.go
jpogran: I don't understand the reference. I stole/copied this approach from the `snippetForExprContraints` method

decoder/label_candidates.go
jpogran: Once I figure out how to model it, I will 😁

decoder/label_schema_test.go
jpogran: Done!

decoder/expression_candidates.go
radeksimko: What I meant is that you can call `snippetForExprContraints` for the inner element of each complex type (SetExpr, ListExpr, first element of TupleExpr & TupleConsExpr). This will lead to e.g.
```hcl
required_attribute = [ "|" ]
```
as opposed to
```hcl
required_attribute = [ | ]
```
(assuming the inner element is of string type)

It's minor thing though - so feel free to ignore! 😉

radeksimko: FWIW there's some planned work to generally build support for arbitrarily nested types, so we can address it as part of that task: https://github.com/hashicorp/terraform-ls/issues/496

schema/body_schema.go
jpogran: I've decided to leave them where they are for now to not complicate things further for this PR

PR #94, opened 2021-11-08
Introduce LocalOrigin + PathOrigin & schema.PathTarget

radeksimko:
This PR introduces a concept of a path-based origin, i.e. origin reference that is pointing to a different path, in Terraform that would be outside of a module.

In Terraform this can be represented by a module input name, pointing to the relevant variable, e.g.

```hcl
module "example" {
  source = "./module"
  foo = "bar"
}
```
where `foo` would be the `PathOrigin` and point to this block within `./module/`
```hcl
variable "foo" {
  type = string
}
```

Relatedly all the existing origins become `LocalOrigin` with all the same existing structure.

There were a few high-level API changes in that both reference-lookup-related methods were moved under `Decoder` (from `PathDecoder`) as they can now both return results outside the context of a single path and therefore also need access to other paths via the `PathReader`.

It's not the smallest PR, but also it's not as big as some other recent ones 😅  and hopefully the commit log can be helpful in walking through it.

### DIFF
--- a/decoder/code_lens.go
+++ b/decoder/code_lens.go
@@ -11,19 +11,25 @@ import (
 // and returns any errors from all of them.
 // i.e. Error from an earlier lens doesn't prevent future lens
 // from being executed
-func (d *PathDecoder) CodeLensesForFile(ctx context.Context, file string) ([]lang.CodeLens, error) {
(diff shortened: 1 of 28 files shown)


dbanck (approved):
I was able to follow the changes in each commit and think everything makes sense so far. However, I cannot provide a "real" approval (yet) because everything is still new, and I can't oversee all implications.

PR #102, opened 2022-03-02
Improve accuracy of reported semantic tokens for type declarations

radeksimko:
This is to support https://github.com/hashicorp/terraform-ls/issues/816 by providing semantic tokens which more accurately represent valid type declarations.

This essentially avoids reporting incorrect type declarations such as

```hcl
variable "example" {
  type = foobar
}
variable "example2" {
  type = list(blah)
}
```

We could of course go further by upstreaming some of the logic into https://github.com/hashicorp/hcl/tree/main/ext/typeexpr, and/or further improve accuracy by reflecting that e.g. collection types expect only primitive types inside and map/object types expect `{}` inside, but that goes beyond the "low hanging fruit" I'm trying to address here and would fall more broadly under https://github.com/hashicorp/terraform-ls/issues/496


dbanck (approved):
LGTM 👍

PR #105, opened 2022-03-14
fix: filtering of self references for cross-file references

radeksimko:
As demonstrated by the attached test this fixes a case of completion of references where we intend to filter out external self-references to the outer block itself when the completion is requested within the block. Previously we'd only be comparing the range of completion and reference, without considering the filename.

Related: https://github.com/hashicorp/terraform-ls/issues/646


dbanck (approved):
This makes sense! Good catch

PR #106, opened 2022-03-15
Support custom modifiers via `BlockSchema` & `LabelSchema`

radeksimko:
The LSP spec is not explicit in saying that custom token types or modifiers are supported, but it does neither explicitly say it's forbidden.

The fact that the [underlying type for both is a `string`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokenModifiers) and that there is relatively detailed corresponding client/server capability available suggests this is possible.

The expectation is that downstream LS (`terraform-ls`) would use the client capability to see what token modifiers the client actually supports before offering them any custom ones and it would continue providing fallback, just like terraform-ls does today:

https://github.com/hashicorp/terraform-ls/blob/2de83c81f9e53457b943efaf0801acf6185b71b3/internal/lsp/token_encoder.go#L37-L86

The benefit is that clients which do support more specific token types and modifiers can use those - which in turn enables more specific themes that can take advantage of these. Clients which do not support these will continue to use the fallback types & modifiers. So basically - everyone's happy? 😄  🤞🏻 

Related: https://github.com/hashicorp/vscode-terraform/pull/958


radeksimko:
> Do you think it would be useful to have semantic token modifiers on attributes, too?

Potentially yes, although I imagine traversals would be wanted more. I wanted to ship at least basic support for this though. We can always add more later, removing is typically harder once people rely on it and I'm not entirely sure what the best modifiers would be in the context of traversals and attributes yet.

We already set dependent modifier for some attributes, like `module` `source` or `resource` `provider`, which folks may want to highlight differently. I'm guessing some might want all core/meta-attributes to be highlighted differently? but I don't know that for sure.

Similarly for traversals folks may want to highlight the whole address (e.g. `data.aws_instance.name`) depending on what logical block it belongs to - but I don't know for sure. Some folks wanted to highlight the first "root" segment - e.g. `data` in `data.aws_instance.name` but that leaves question marks for resource addresses which do not have any such dedicated segment.

So ... TL;DR lots of questions to answer still IMO.

dbanck:
Thanks for the explanation and for raising more questions 😄

PR #109, opened 2022-03-21
fix: avoid semantic token modifier inheritance for dependent bodies

### LINKED ISSUE #108: Limit block type semantic token modifier inheritance to `BodySchema`
## Background

Currently we let the block type modifiers be inherited indefinitely to any nested blocks within.

I am not sure whether that's what people want, but it's not something I originally intended in my initial PR.

This currently results in all nested blocks to have applied the same style as the root one:

Screenshot 2022-03-17 at 16 27 07

## Proposal

Avoid inheritance for `DependentBody` body schemas.

radeksimko:
Fixes #108 

## UX

Nested dependent blocks no longer inherit the top level modifiers.

### Before

Screenshot 2022-03-21 at 16 46 33

### After

Screenshot 2022-03-21 at 16 44 22

The examples used a custom VS Code theme:

```json
{
  "name": "Example Theme",
  "semanticHighlighting": true,
  "semanticTokenColors": {
    "hcl-blockType.terraform-resource": {"fontStyle": "bold", "foreground": "#fff"},
    "hcl-blockType.terraform-data": {"fontStyle": "bold", "foreground": "#fff"},
    "hcl-blockType.terraform-module": {"fontStyle": "bold", "foreground": "#3d5aff"},
    "hcl-blockLabel.terraform-module.terraform-name": {"foreground": "#bac4ff"},
    "hcl-blockLabel.terraform-type": {"foreground": "#844FBA", "fontStyle": "bold"},
    "hcl-blockLabel.terraform-resource.terraform-name": {"foreground": "#a589b3"}
  }
}
```

### DIFF
--- a/decoder/semantic_tokens.go
+++ b/decoder/semantic_tokens.go
@@ -127,7 +127,7 @@ func (d *PathDecoder) tokensForBody(body *hclsyntax.Body, bodySchema *schema.Bod
 
 		depSchema, _, ok := NewBlockSchema(blockSchema).DependentBodySchema(block.AsHCLBlock())
 		if ok {
-			tokens = append(tokens, d.tokensForBody(block.Body, depSchema, true, blockModifiers)...)
+			tokens = append(tokens, d.tokensForBody(block.Body, depSchema, true, []lang.SemanticTokenModifier{})...)
 		}
 	}
 


dbanck (approved):
LGTM!

How to we want to highlight "normal" nested blocks inside a body? For example, `lifecycle`:

CleanShot 2022-03-22 at 12 25 24@2x

radeksimko:
> How to we want to highlight "normal" nested blocks inside a body?

Frankly I don't know _yet_. My plan was to land it like this and then solicit feedback from theme authors. One direction would be the one described already in https://github.com/hashicorp/terraform-ls/issues/835 - i.e. Core vs Provider blocks

https://github.com/hashicorp/hcl-lang/pull/110 should give us some more flexibility downstream in `terraform-schema` and react to various needs.

PR #110, opened 2022-03-22
schema: Make modifiers plural in LabelSchema

radeksimko:
This is to enable https://github.com/hashicorp/terraform-schema/issues/96

---

The PR introduces a number of changes:

 - `SemanticTokenModifier` in `LabelSchema` becomes plural, allowing assignment of more than 1 modifier via schema
 - `SemanticTokenModifier` in `BlockSchema` becomes plural, allowing assignment of more than 1 modifier via schema (this actually remains unused downstream in https://github.com/hashicorp/terraform-schema/pull/100 but it's mostly just to align the API and make it more flexible 🤷🏻 
 - Introduce `SemanticTokenModifiers` to `AttrSchema` such that we can use it downstream to report dependent attributes and later for adding any more modifiers per https://github.com/hashicorp/terraform-ls/issues/835
 - Removes auto-assignment of the dependent modifier, to apply the "principle of least surprise". https://github.com/hashicorp/terraform-schema/pull/100 basically reintroduces these modifiers more explicitly - which can be verbose but I still believe explicit & verbose is better than brief and magic.

### DIFF
--- a/decoder/dependent_body_test.go
+++ b/decoder/dependent_body_test.go
@@ -111,17 +111,22 @@ func TestBodySchema_DependentBodySchema_dependentAttr(t *testing.T) {
 	firstDepBody := &schema.BodySchema{
 		Attributes: map[string]*schema.AttributeSchema{
 			"backend": {
-				Expr:     schema.LiteralTypeOnly(cty.String),
-				IsDepKey: true,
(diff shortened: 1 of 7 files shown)


dbanck (approved):
LGTM

Thanks for splitting this into four easy to follow commits

Issue #101, opened 2022-01-27
Sort `decoder.ReferenceOrigins` before returning from `ReferenceOriginsTargetingPos`


radeksimko:
As caught in CI tests: https://github.com/hashicorp/hcl-lang/runs/4962645944?check_suite_focus=true#step:10:468

```
=== RUN   TestReferenceOriginsTargetingPos/8-matching_local_and_path_origin
    reference_origins_test.go:758: mismatch of reference origins:   decoder.ReferenceOrigins{
        - 	{
        - 		Path:  lang.Path{Path: "/tmp/TestReferenceOriginsTargetingPos3177626092/001/alpha"},
        - 		Range: s"test.tf:6,20-27",
        - 	},
          	{Path: {Path: "/tmp/TestReferenceOriginsTargetingPos3177626092/001/beta"}, Range: {Filename: "test.tf", Start: {Line: 1, Column: 10, Byte: 9}, End: {Line: 1, Column: 17, Byte: 16}}},
        + 	{
        + 		Path:  lang.Path{Path: "/tmp/TestReferenceOriginsTargetingPos3177626092/001/alpha"},
        + 		Range: s"test.tf:6,20-27",
        + 	},
          }
--- FAIL: TestReferenceOriginsTargetingPos (0.00s)
```

PR #111, opened 2022-04-01
Fix reference collection sorting

dbanck:
This PR fixes a small typo within the sorting function for the collected references


radeksimko (approved):
Thanks for the catch!

I guess this closes https://github.com/hashicorp/hcl-lang/issues/101 then?

dbanck:
> I guess this closes #101 then?

No, I think the failing tests in #101 are all testing the `ReferenceOriginsTargetingPos` function.

This fix is inside `CollectReferenceOrigins` which is called by the language server as part of the module decoding.

PR #112, opened 2022-04-01
Expose new kind of origins

dbanck:
This PR allows schemas to introduce a target for the module source attribute. We achieve this by introducing a new field `Targets` to the `BodySchema`.

Example within terraform-schema: https://github.com/hashicorp/terraform-schema/pull/101

Required for https://github.com/hashicorp/terraform-ls/issues/558

### DIFF
--- a/decoder/reference_origins.go
+++ b/decoder/reference_origins.go
@@ -111,6 +111,14 @@ func (d *PathDecoder) referenceOriginsInBody(body hcl.Body, bodySchema *schema.B
 			}
 		}
 
+		if aSchema.IsDepKey && bodySchema.Targets != nil {
+			origins = append(origins, reference.DirectOrigin{
+				Range:       attr.Expr.Range(),
+				TargetPath:  bodySchema.Targets.Path,
+				TargetRange: bodySchema.Targets.Range,
+			})
+		}
+
 		origins = append(origins, d.findOriginsInExpression(attr.Expr, aSchema.Expr)...)
 	}
 
--- a/reference/origin_direct.go
+++ b/reference/origin_direct.go
@@ -1,0 +1,35 @@
+package reference
+
+import (
+	"github.com/hashicorp/hcl-lang/lang"
+	"github.com/hashicorp/hcl/v2"
+)
+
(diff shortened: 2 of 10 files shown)


decoder/reference_origins.go
```diff
@@ -111,6 +111,14 @@ func (d *PathDecoder) referenceOriginsInBody(body hcl.Body, bodySchema *schema.B
 			}
 		}
 
+		if aSchema.IsDepKey && bodySchema.LinksTo != nil {
```
dbanck, line 114: It _feels_ like this is not guarded enough and might introduce `DirectOrigin`s at unwanted locations.

schema/body_schema.go
```diff
@@ -31,13 +31,22 @@ type BodySchema struct {
 	// TargetableAs represents how else the body may be targeted
 	// if not by its declarable attributes or blocks.
 	TargetableAs Targetables
+
+	// LinksTo indicates whether the something inside the body is
+	// linking to another location
+	LinksTo *LinksTo
```
dbanck, line 37: I'm very open to better naming suggetions

decoder/reference_origins.go
dbanck: Furthermore, all origins would link to the same location, and we don't support having multiple links inside a body.

reference/origin_direct.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/hashicorp/hcl/v2"
+)
+
+// DirectOrigin represents an origin which directly targets a file
+// and doesn't need a matching target
+type DirectOrigin struct {
+	// Range represents a range of a local traversal or an attribute
+	Range hcl.Range
+
+	// TargetPath represents what Path does the origin targets
```
radeksimko, line 14:
```suggestion
	// TargetPath represents what (directory) Path does the origin targets
```
Nitpick - just to clarify this is not a file path

radeksimko (approved):
Thank you for all the tests and comments.

LGTM - I just left you some very minor in-line suggestions/comments, but none are blocking.

radeksimko (approved):
:shipit:

PR #114, opened 2022-04-12
ci: reduce min threshold for k8s-metrics-server benchmark

radeksimko:
This is to address the following nightly failure:

https://github.com/hashicorp/terraform-ls/runs/5983427527?check_suite_focus=true#step:9:1

```json
{
  "Status": "fail",
  "Diffs": [
    {
      "Status": "fail",
      "Package": "github.com/hashicorp/terraform-ls/internal/langserver/handlers",
      "Benchmark": "BenchmarkInitializeFolder_basic/k8s-metrics-server-2",
      "Value": 1057.225971
    }
  ],
  "Thresholds": {
    "Min": 1060,
    "Max": 2800
  }
}
```

### DIFF
--- a/decoder/candidates.go
+++ b/decoder/candidates.go
@@ -13,7 +13,7 @@ import (
 //
 // Schema is required in order to return any candidates and method will return
 // error if there isn't one.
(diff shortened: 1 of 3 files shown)


radeksimko:
Wrong window 🤦🏻

PR #115, opened 2022-04-12
Enable DocsLink for attributes

dbanck:
Required for https://github.com/hashicorp/vscode-terraform/issues/673

This PR extends the collection of links inside a file from just blocks to attributes. With the change it's now possible to have an attribute (like `source` inside a module block) link to documentation.

### DIFF
--- a/decoder/links.go
+++ b/decoder/links.go
@@ -44,16 +44,24 @@ func (d *PathDecoder) linksInBody(body *hclsyntax.Body, bodySchema *schema.BodyS
 		if block.Body != nil {
 			depSchema, dk, ok := NewBlockSchema(blockSchema).DependentBodySchema(block.AsHCLBlock())
 			if ok && depSchema.DocsLink != nil {
+				link := depSchema.DocsLink
+				u, err := d.docsURL(link.URL, "documentLink")
+				if err != nil {
+					continue
+				}
 				for _, labelDep := range dk.Labels {
-					link := depSchema.DocsLink
-					u, err := d.docsURL(link.URL, "documentLink")
-					if err == nil {
-						links = append(links, lang.Link{
-							URI:     u.String(),
-							Tooltip: link.Tooltip,
-							Range:   block.LabelRanges[labelDep.Index],
-						})
-					}
+					links = append(links, lang.Link{
+						URI:     u.String(),
+						Tooltip: link.Tooltip,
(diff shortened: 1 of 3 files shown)


decoder/links.go
```diff
@@ -56,11 +75,16 @@ func (d *PathDecoder) linksInBody(body *hclsyntax.Body, bodySchema *schema.BodyS
 					}
 				}
```
radeksimko, line 76: I didn't think of this before when we discussed this earlier but could we not range here over [`dk.Attributes`](https://github.com/hashicorp/hcl-lang/blob/f24c834260e015170a294693b9971e4ff1c219ab/schema/dependent_schema.go#L19) here, instead of merging dependent schemas below? 🤔 

That would also avoid the need to deal with attributes above.

dbanck: I missed that `dk` also has an attributes field. Ranging over this makes things much easier, thanks!

decoder/links_test.go
```diff
@@ -118,8 +118,8 @@ func TestLinksInFile_json(t *testing.T) {
 	})
 
 	_, err := d.LinksInFile("test.tf.json")
-	unknownFormatErr := &UnknownFileFormatError{}
-	if !errors.As(err, &unknownFormatErr) {
-		t.Fatal("expected UnknownFileFormatError for JSON body")
+	noSchemaErr := &NoSchemaError{}
+	if !errors.As(err, &noSchemaErr) {
+		t.Fatal("expected NoSchemaError for JSON body")
```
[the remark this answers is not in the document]
dbanck: This update was the result of the usage of `decodeBody` in the links function. With a now much simpler implementation, I've reverted the changes an added a short comment.

decoder/links_test.go
```diff
@@ -101,6 +101,92 @@ func TestLinksInFile(t *testing.T) {
...
+				"num_attr": {Expr: schema.LiteralTypeOnly(cty.Number)},
+				"source": {
+					Expr:        schema.LiteralTypeOnly(cty.String),
+					Description: lang.PlainText("Special attribute"),
+					IsDepKey:    true,
+				},
+			},
+		},
+		DependentBody: map[schema.SchemaKey]*schema.BodySchema{
+
```
radeksimko, line 122:
```suggestion
```

radeksimko (approved):
LGTM 🚀

PR #122, opened 2022-07-15
Enable go to module output

dbanck:
WIP


schema/body_schema.go
```diff
@@ -32,9 +32,34 @@ type BodySchema struct {
...
-	// dependent on an attribute (e.g. Terraform module source)
+	// Targets represent a location targeted by the body; when used as a body
+	// dependent on an attribute (e.g., Terraform module source)
 	Targets *Target
+
+	// AdditionalOrigins represent a list of origins we should revisit during
+	// reference origin collection. For example, module outputs can be
+	// referenced from still unknown locations during the build of the module
+	// schema.
+	AdditionalOrigins ImpliedOrigins
```
radeksimko, line 43: Nitpick: Should we call this `ImpliedOrigins` instead of `AdditionalOrigins`, to communicate the fact that these are not regular reference origins?

radeksimko (approved):
LGTM, just one minor question about naming.

schema/body_schema.go
dbanck: Good suggestion! I've been thinking of this, too. Currently, we use both terms for the same thing — which might lead to confusion. I'll rename the field.

PR #123, opened 2022-07-27
Enable hooks for dynamic completion

dbanck:
This PR extends the completion of attribute values by getting completion candidates from dynamically registered hooks.

If an attribute schema contains one (or multiple) completion hooks, we look in the global decoder context to see if those hooks are registered. If they are, each hook is executed and should return a list of completion candidates.

The resulting candidate always replaces the whole attribute value range (including quotes).

---

* Part of https://github.com/hashicorp/terraform-ls/issues/868

### DIFF
--- a/decoder/completion_resolve.go
+++ b/decoder/completion_resolve.go
@@ -2,9 +2,16 @@ package decoder
 
 import (
 	"context"
+	"errors"
 )
 
+// ResolveCandidate tries to gather more information for a candidate,
+// by checking for a resolve hook and executing it.
 func (d *Decoder) ResolveCandidate(ctx context.Context, unresolvedCandidate UnresolvedCandidate) (*ResolvedCandidate, error) {
+	if unresolvedCandidate.ResolveHook == nil {
+		return nil, errors.New("missing resolve hook")
+	}
+
 	if resolveFunc, ok := d.ctx.CompletionResolveHooks[unresolvedCandidate.ResolveHook.Name]; ok {
 		return resolveFunc(ctx, unresolvedCandidate)
 	}
--- a/decoder/completion_resolve.go
+++ b/decoder/completion_resolve.go
@@ -1,0 +1,13 @@
+package decoder
+
+import (
+	"context"
+)
+
+func (d *Decoder) ResolveCandidate(ctx context.Context, unresolvedCandidate UnresolvedCandidate) (*ResolvedCandidate, error) {
+	if resolveFunc, ok := d.ctx.CompletionResolveHooks[unresolvedCandidate.ResolveHook.Name]; ok {
+		return resolveFunc(ctx, unresolvedCandidate)
+	}
+
+	return nil, nil
+}
--- a/decoder/completion_resolve.go
+++ b/decoder/completion_resolve.go
@@ -2,7 +2,6 @@ package decoder
 
 import (
 	"context"
-	"errors"
 )
 
 // ResolveCandidate gathers more information for a completion candidate
--- a/decoder/completion_resolve.go
+++ b/decoder/completion_resolve.go
@@ -5,11 +5,12 @@ import (
 	"errors"
 )
 
-// ResolveCandidate tries to gather more information for a candidate,
+// ResolveCandidate gathers more information for a completion candidate
 // by checking for a resolve hook and executing it.
+// This would be called as part of `completionItem/resolve` LSP method.
 func (d *Decoder) ResolveCandidate(ctx context.Context, unresolvedCandidate UnresolvedCandidate) (*ResolvedCandidate, error) {
 	if unresolvedCandidate.ResolveHook == nil {
-		return nil, errors.New("missing resolve hook")
+		return nil, nil
 	}
 
 	if resolveFunc, ok := d.ctx.CompletionResolveHooks[unresolvedCandidate.ResolveHook.Name]; ok {
--- a/decoder/context.go
+++ b/decoder/context.go
@@ -28,6 +28,7 @@ type DecoderContext struct {
 	// attribute schema, one can refer to the hooks map key to enable the hook
 	// execution on CompletionAtPos.
 	CompletionHooks CompletionFuncMap
+
 	// CompletionResolveHooks represents a map of available hooks for
 	// completion candidate resolving. One can register new hooks by adding an
 	// entry to this map. On completion candidate creation, one can specify a
@@ -48,11 +49,18 @@ func (d *Decoder) SetContext(ctx DecoderContext) {
 	d.ctx = ctx
 }
 
+// CompletionFunc is the function signature for completion hooks.
+//
+// The completion func has access to path, filename and pos via context:
+//  path, ok := decoder.PathFromContext(ctx)
+//  filename, ok := decoder.FilenameFromContext(ctx)
+//  pos, ok := decoder.PosFromContext(ctx)
 type CompletionFunc func(ctx context.Context, value cty.Value) ([]Candidate, error)
 type CompletionFuncMap map[string]CompletionFunc
 
-type CompletionResolveFuncMap map[string]CompletionResolveFunc
+// CompletionResolveFunc is the function signature for resolve hooks.
 type CompletionResolveFunc func(ctx context.Context, unresolvedCandidate UnresolvedCandidate) (*ResolvedCandidate, error)
+type CompletionResolveFuncMap map[string]CompletionResolveFunc
 
 // Candidate represents a completion candidate created and returned from a
 // completion hook.
--- a/decoder/context.go
+++ b/decoder/context.go
@@ -32,8 +32,8 @@ type DecoderContext struct {
 	// completion candidate resolving. One can register new hooks by adding an
 	// entry to this map. On completion candidate creation, one can specify a
 	// resolve hook by using the map key string. If a completion candidate has
-	// a resolve hook, ResolveCandidate will execute the hook and update the
-	// completion item.
+	// a resolve hook, ResolveCandidate will execute the hook and return
+	// additional (resolved) data for the completion item.
 	CompletionResolveHooks CompletionResolveFuncMap
 }
 
--- a/decoder/context.go
+++ b/decoder/context.go
@@ -2,6 +2,7 @@ package decoder
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/hashicorp/hcl-lang/lang"
 	"github.com/zclconf/go-cty/cty"
@@ -22,8 +23,25 @@ type DecoderContext struct {
 	// which will be executed in the exact order they're declared
 	CodeLenses []lang.CodeLensFunc
 
-	CompletionHooks        CompletionFuncMap
+	// CompletionHooks represents a map of available hooks for completion.
+	// One can register new hooks by adding an entry to this map. Inside the
+	// attribute schema, one can refer to the hooks map key to enable the hook
+	// execution on CompletionAtPos.
+	CompletionHooks CompletionFuncMap
+	// CompletionResolveHooks represents a map of available hooks for
+	// completion candidate resolving. One can register new hooks by adding an
+	// entry to this map. On completion candidate creation, one can specify a
+	// resolve hook by using the map key string. If a completion candidate has
+	// a resolve hook, ResolveCandidate will execute the hook and update the
+	// completion item.
 	CompletionResolveHooks CompletionResolveFuncMap
+}
+
+func NewDecoderContext() DecoderContext {
+	return DecoderContext{
+		CompletionHooks:        make(CompletionFuncMap),
+		CompletionResolveHooks: make(CompletionResolveFuncMap),
+	}
 }
 
 func (d *Decoder) SetContext(ctx DecoderContext) {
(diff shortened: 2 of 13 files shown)


decoder/expression_candidates.go
```diff
@@ -152,15 +165,103 @@ func constraintsAtPos(expr hcl.Expression, constraints ExprConstraints, pos hcl.
...
+		if completionFunc, ok := d.decoderCtx.CompletionHooks[hook.Name]; ok {
+			res, _ := completionFunc(ctx, cty.StringVal(prefix))
+
+			for _, c := range res {
+				// We're adding quotes to the string here, as we're always
+				// replacing the whole edit range
+				text := fmt.Sprintf("%q", c.RawInsertText)
+
+				candidates = append(candidates, lang.Candidate{
+					Label:        text,
```
dbanck, line 247: We're using `text` instead of `c.Label` here, because VS Code is using the label value to filter the text edits. So they need to match. I could look into setting `filterText` on `lang.Candidate` instead 🤔

decoder/completion_resolve.go
```diff
@@ -0,0 +1,13 @@
+package decoder
+
+import (
+	"context"
+)
+
+func (d *Decoder) ResolveCandidate(ctx context.Context, unresolvedCandidate UnresolvedCandidate) (*ResolvedCandidate, error) {
```
radeksimko, line 7: Nitpick: Do you mind adding a comment to the method, as it is part of the public API?

decoder/completion_resolve.go
```diff
@@ -0,0 +1,13 @@
...
+	if resolveFunc, ok := d.ctx.CompletionResolveHooks[unresolvedCandidate.ResolveHook.Name]; ok {
```
radeksimko, line 8: Do you mind adding a `nil` check here for `unresolvedCandidate.ResolveHook` and return early, to avoid crashes?

decoder/context.go
```diff
@@ -16,8 +21,79 @@ type DecoderContext struct {
 	// CodeLenses represents a slice of executable lenses
 	// which will be executed in the exact order they're declared
 	CodeLenses []lang.CodeLensFunc
+
+	CompletionHooks        CompletionFuncMap
+	CompletionResolveHooks CompletionResolveFuncMap
```
radeksimko, line 26: Can you just briefly explain in comments above each field how these work? i.e. that this is place to register hooks, which makes them available for the schema and based on the schemas, they get looked up during `CompletionAtPos()` and `ResolveCandidate()` based on the map key - or something along these lines.

decoder/context.go
```diff
@@ -16,8 +21,79 @@ type DecoderContext struct {
...
 	d.ctx = ctx
 }
+
+type CompletionFunc func(ctx context.Context, value cty.Value) ([]Candidate, error)
+type CompletionFuncMap map[string]CompletionFunc
+
+type CompletionResolveFuncMap map[string]CompletionResolveFunc
+type CompletionResolveFunc func(ctx context.Context, unresolvedCandidate UnresolvedCandidate) (*ResolvedCandidate, error)
+
+type Candidate struct {
```
radeksimko, line 39: I'm half wondering if we should call this `HookCandidate`, but given that we don't have a conflict with another "candidate" name within the package I'm fine with this.

Either way though we should document this via comment to clarify that this represents completion candidates returned from hooks.

decoder/context.go
```diff
@@ -16,8 +21,79 @@ type DecoderContext struct {
...
+	Kind lang.CandidateKind
+
+	// Description represents human-readable description
+	// of the candidate
+	Description lang.MarkupContent
+
+	// IsDeprecated indicates whether the candidate is deprecated
+	IsDeprecated bool
+
+	RawInsertText string
```
radeksimko, line 57: Do you mind adding a comment to briefly explain what it's for and how we treat the value in terms of quoting - i.e. that we expect it to be already quoted, if necessary? Just so that consumers know and don't have to find out via trial & error or by studying the implementation 😅

decoder/decoder_test.go
```diff
@@ -41,6 +41,10 @@ func testPathDecoder(t *testing.T, pathCtx *PathContext) *PathDecoder {
 	d := NewDecoder(&testPathReader{
 		paths: dirs,
 	})
+	d.SetContext(DecoderContext{
```
radeksimko, line 44: A common convention is that if the struct isn't safe to use directly (i.e. one has to allocate maps or slices because the underlying code doesn't do it lazily), it has a constructor, such as `NewDecoderContext()`.

So I think we should either do the lazy allocation, and/or check for `nil`s where we may be accessing unallocated maps, or create and use the constructor.

decoder/expression_candidates.go
radeksimko: I think this quoting logic should live in the helper which converts `decoder.ExpressionCandidate` to `decoder.Candidate` as this would not be relevant in any context other than if we're completing (string) expressions. I know there's currently no way we would complete anything, but it would be good to be prepared for that eventuality.

So `Label` can then stay set to `c.Label` here. The same goes for `RawInsertText` - I think the main reason we picked that name was to convey that it gets inserted "as-is" without side effects, so we should move the quoting of the text edit also.

decoder/expression_candidates_test.go
```diff
@@ -2745,3 +2749,178 @@ variable "ccc" {}
...
+		},
+	}
+
+	for i, tc := range testCases {
+		t.Run(fmt.Sprintf("%d-%s", i, tc.testName), func(t *testing.T) {
+			bodySchema := &schema.BodySchema{
+				Attributes: tc.attrSchema,
+			}
+
+			f, _ := hclsyntax.ParseConfig([]byte(tc.cfg), "test.tf", hcl.InitialPos)
```
radeksimko, line 2905: Nitpick: A comment to explain why we're ignoring the diagnostics here would be useful (i.e. that we expect invalid configs), just to decrease the chance of someone just blindly copy pasting it elsewhere where this context is not applicable.

lang/candidate.go
```diff
@@ -42,6 +42,8 @@ type Candidate struct {
 	// TriggerSuggest allows server to instruct the client whether
 	// to reopen candidate suggestion popup after insertion
 	TriggerSuggest bool
+
+	ResolveHook *ResolveHook
```
radeksimko, line 46: Similar to above - do you mind adding a brief comment to explain the purpose of that field?

schema/attribute_schema.go
```diff
@@ -36,6 +36,10 @@ type AttributeSchema struct {
 	// to report for the attribute name
 	// (in addition to any modifiers of any parent blocks)
 	SemanticTokenModifiers lang.SemanticTokenModifiers
+
+	// CompletionHooks represents any hooks which provide
+	// additional completion candidates for the attribute dynamically
```
radeksimko, line 41:
```suggestion
	// additional completion candidates for the attribute.
	// These are typically candidates which cannot be provided
	// via schema and come from external APIs or other sources.
```

schema/body_schema.go
```diff
@@ -12,8 +12,11 @@ import (
 // BodySchema describes schema of a body comprised of blocks or attributes
 // (if any), where body can be root or body of any block in the hierarchy.
 type BodySchema struct {
-	Blocks       map[string]*BlockSchema
-	Attributes   map[string]*AttributeSchema
+	Blocks     map[string]*BlockSchema
+	Attributes map[string]*AttributeSchema
+	// AnyAttribute represents an attribute where a user can pick any arbitrary
```
radeksimko, line 17:
```suggestion

	// AnyAttribute represents an attribute where a user can pick any arbitrary
```

decoder/context.go
```diff
@@ -16,8 +21,79 @@ type DecoderContext struct {
...
+
+	// ResolveHook represents a resolve hook to call
+	// and any arguments to pass to it
+	ResolveHook *lang.ResolveHook
+}
+
+type ExpressionCandidate struct {
```
radeksimko, line 64: Do you mind adding a comment to this, to indicate that this is the preferred way to provide completion candidates for attributes values (expressions) and to note that there's `ExpressionCompletionCandidate` to convert it into the low-level `Candidate`?

radeksimko (commented):
I left some comments in-line, mostly minor though.

Thank you for getting to the bottom of the quoting issue!

All in all it looks pretty good and I guess we can tweak the detection of empty expressions later, when/if necessary, based on some real configs which we find to trigger any bugs.

decoder/expression_candidates_test.go
dbanck: I just blindly copy pasted it from above 😇

decoder/completion_resolve.go
```diff
@@ -0,0 +1,20 @@
+package decoder
+
+import (
+	"context"
+	"errors"
+)
+
+// ResolveCandidate tries to gather more information for a candidate,
+// by checking for a resolve hook and executing it.
```
radeksimko, line 9:
```suggestion
// ResolveCandidate gathers more information for a completion candidate
// by checking for a resolve hook and executing it.
// This would be called as part of `completionItem/resolve` LSP method.
```

decoder/completion_resolve.go
```diff
@@ -0,0 +1,20 @@
...
+func (d *Decoder) ResolveCandidate(ctx context.Context, unresolvedCandidate UnresolvedCandidate) (*ResolvedCandidate, error) {
+	if unresolvedCandidate.ResolveHook == nil {
+		return nil, errors.New("missing resolve hook")
```
radeksimko, line 12:
```suggestion
		return nil, nil
```
We could of course ignore that error downstream in LS or assume extra check for `nil` there, but I'm not sure there's any benefit of raising that error in the first place? Typically we'd return errors to make testing easier, but I think this is testable even without the error?

decoder/context.go
```diff
@@ -16,8 +22,117 @@ type DecoderContext struct {
 	// CodeLenses represents a slice of executable lenses
 	// which will be executed in the exact order they're declared
 	CodeLenses []lang.CodeLensFunc
+
+	// CompletionHooks represents a map of available hooks for completion.
+	// One can register new hooks by adding an entry to this map. Inside the
+	// attribute schema, one can refer to the hooks map key to enable the hook
+	// execution on CompletionAtPos.
+	CompletionHooks CompletionFuncMap
```
radeksimko, line 30:
```suggestion
	CompletionHooks CompletionFuncMap

```

decoder/context.go
```diff
@@ -16,8 +22,117 @@ type DecoderContext struct {
...
+	// CompletionResolveHooks represents a map of available hooks for
+	// completion candidate resolving. One can register new hooks by adding an
+	// entry to this map. On completion candidate creation, one can specify a
+	// resolve hook by using the map key string. If a completion candidate has
+	// a resolve hook, ResolveCandidate will execute the hook and update the
+	// completion item.
```
radeksimko, line 36:
```suggestion
	// a resolve hook, ResolveCandidate will execute the hook and return
	// additional (resolved) data for the completion item.
```

lang/candidate.go
```diff
@@ -42,6 +42,10 @@ type Candidate struct {
 	// TriggerSuggest allows server to instruct the client whether
 	// to reopen candidate suggestion popup after insertion
 	TriggerSuggest bool
+
+	// ResolveHook allows the async resolution of additional information
```
radeksimko, line 46:
```suggestion
	// ResolveHook allows resolution of additional information
```
I understand the intentions here, but I think it could be potentially misunderstood for protocol-level synchronicity in the context of this library, which is in fact a concern of the consumer (LS) and we don't use any sync primitives in `hcl-lang` really.

schema/attribute_schema.go
```diff
@@ -36,6 +36,12 @@ type AttributeSchema struct {
 	// to report for the attribute name
 	// (in addition to any modifiers of any parent blocks)
 	SemanticTokenModifiers lang.SemanticTokenModifiers
+
+	// CompletionHooks represents any hooks which provide
```
radeksimko, line 40:
```suggestion
	// CompletionHooks represent any hooks which provide
```

decoder/context.go
```diff
@@ -16,8 +22,117 @@ type DecoderContext struct {
...
+		CompletionHooks:        make(CompletionFuncMap),
+		CompletionResolveHooks: make(CompletionResolveFuncMap),
+	}
 }
 
 func (d *Decoder) SetContext(ctx DecoderContext) {
 	d.ctx = ctx
 }
+
+type CompletionFunc func(ctx context.Context, value cty.Value) ([]Candidate, error)
```
radeksimko, line 51: As mentioned in the other PR review
 - https://github.com/hashicorp/terraform-ls/pull/1017#discussion_r933310836
 - https://github.com/hashicorp/terraform-ls/pull/1017#discussion_r933313834

is it worth bringing some of the docs here, so we can omit them downstream?

radeksimko (approved):
LGTM, just some really minor suggestions in-line. 👍🏻

decoder/completion_resolve.go
radeksimko: @dbanck I think this still remains unresolved - and would lead to a crash if passed `unresolvedCandidate.ResolveHook` is `nil`.

decoder/completion_resolve.go
radeksimko: @dbanck I think this still remains unresolved - and would lead to a crash if passed `unresolvedCandidate.ResolveHook` is `nil`.

radeksimko: Actually - ignore me - somehow I was looking at an old revision! 🙈

decoder/context.go
dbanck: That's a valid point, I moved it here.

I'm just wondering how easy the docs are discoverable, one won't be able to see them anywhere on hover in the LS.

radeksimko: You could add a reference or comment to `CompletionFuncMap` which is easier to discover downstream on hover.

PR #124, opened 2022-07-27
Sort reference origins

dbanck:
After running into this a couple of times [recently](https://github.com/hashicorp/hcl-lang/runs/7543556000?check_suite_focus=true), this PR now sorts the reference origins by path, filename and range.

Closes https://github.com/hashicorp/hcl-lang/issues/101


radeksimko (approved):
Thanks! :shipit:

radeksimko (changes requested):
Actually the tests still seem to be failing

dbanck:
I shouldn't copy and extend sort functions without checking them first.

It should work now
CleanShot 2022-07-27 at 20 17 54

radeksimko (approved):
Thanks for the update

PR #126, opened 2022-08-03
Add `SortText` and respect `maxCandidates` for hook completion

dbanck:
This PR adds the `SortText` field to completion candidates and only allows `maxCandidates` (100 for now) to be returned on attribute expression completion.

We stop adding candidates and executing hooks once the limit is reached. A better way might be to allow each hook to contribute to the list of completion candidates.

### DIFF
--- a/decoder/expression_candidates.go
+++ b/decoder/expression_candidates.go
@@ -251,6 +251,7 @@ func (d *PathDecoder) candidatesFromHooks(ctx context.Context, attr *hclsyntax.A
 						Range:   editRng,
 					},
 					ResolveHook: c.ResolveHook,
+					SortText:    c.SortText,
 				})
 			}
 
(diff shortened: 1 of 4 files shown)


decoder/expression_candidates.go
```diff
@@ -186,6 +192,13 @@ func FilenameFromContext(ctx context.Context) (string, bool) {
 	return f, ok
 }
 
+type maxCandidatesKey struct{}
+
+func MaxCandidatesFromContext(ctx context.Context) (uint, bool) {
+	f, ok := ctx.Value(maxCandidatesKey{}).(uint)
+	return f, ok
```
radeksimko, line 199:
```suggestion
	mc, ok := ctx.Value(maxCandidatesKey{}).(uint)
	return mc, ok
```

radeksimko (approved):
LGTM! :shipit:

PR #134, opened 2022-09-09
Support count and count.index completion in blocks

jpogran:
This provides completion hints inside blocks for `count.index` references within resource, data and module blocks anywhere the `count` meta-argument is supported. It detects if count is used already and does not suggest duplicates.

Closes https://github.com/hashicorp/terraform-ls/issues/860


decoder/body_candidates.go
```diff
@@ -17,6 +17,18 @@ func (d *PathDecoder) bodySchemaCandidates(body *hclsyntax.Body, schema *schema.
 	candidates := lang.NewCandidates()
 	count := 0
 
+	if schema.Extensions != nil {
+		// check if this schema supports Count attribute
+		if schema.Extensions.Count {
+			countPresent := isAttributePresent(body, "count")
+			// check if Count is already used inside this body, so we don't
+			// suggest a duplicate
+			if !countPresent {
```
dbanck, line 26:
```suggestion
			if _, ok := body.Attributes["count"]; !ok {
```

We can simplify this by checking directly for the `count` attribute and don't need the `isAttributePresent` helper method. Sorry for the wrong tip to use it.

dbanck (commented):
Quick review from my side: LGTM so far, just minor suggestions.

I think we should copy the BodySchema extensions when merging dependent body schemas as well. To be sure to cover every edge-case.
https://github.com/hashicorp/hcl-lang/blob/3406e0cceafaba2d86cdb97cef17e6a76f467efe/decoder/decoder.go#L75-L77

```
mergedSchema.Extensions = depSchema.Extensions.Copy()
```

decoder/semantic_tokens.go
```diff
@@ -37,21 +41,45 @@ func (d *PathDecoder) SemanticTokensInFile(filename string) ([]lang.SemanticToke
...
 		if !ok {
-			if bodySchema.AnyAttribute == nil {
-				// unknown attribute
-				continue
+			if bodySchema.Extensions != nil && name == "count" && bodySchema.Extensions.Count {
+				attrSchema = &schema.AttributeSchema{
+					Description: lang.MarkupContent{
+						Kind:  lang.MarkdownKind,
+						Value: "**count** _optional, number_\n\nThe distinct index number (starting with 0) corresponding to the instance",
+					},
```
radeksimko, line 72:
```suggestion
```
Given that this is purely for semantic tokens, I think we don't need the description here?

decoder/semantic_tokens.go
```diff
@@ -135,6 +165,34 @@ func (d *PathDecoder) tokensForExpression(expr hclsyntax.Expression, constraints
...
+		countIndexAttr := lang.Address{
+			lang.RootStep{
+				Name: "count",
+			},
+			lang.AttrStep{
+				Name: "index",
+			},
+		}
+		countAvailable := icontext.ActiveCountFromContext(ctx)
+		// TODO why is countAvailable not true here?
```
radeksimko, line 178: It _is_ available in the tests you attached and when I was testing it manually?

decoder/semantic_tokens.go
```diff
@@ -135,6 +165,34 @@ func (d *PathDecoder) tokensForExpression(expr hclsyntax.Expression, constraints
...
+			traversal := eType.AsTraversal()
+			tokens = append(tokens, lang.SemanticToken{
+				Type:      lang.TokenTraversalStep,
+				Modifiers: []lang.SemanticTokenModifier{},
+				Range:     traversal[0].SourceRange(),
+			})
+			tokens = append(tokens, lang.SemanticToken{
+				Type:      lang.TokenTraversalStep,
+				Modifiers: []lang.SemanticTokenModifier{},
+				Range:     traversal[1].SourceRange(),
```
radeksimko, line 191: One minor problem with this is that it's misaligned with how we report address segments in all other contexts, that is that we leave out the initial dot.
Screenshot 2022-10-10 at 10 35 20
Screenshot 2022-10-10 at 15 02 36

See LOC 230-235 for how we account for this.

We kinda expect the separator to not have a special meaning in the context of semantic highlighting.

To put it another way, the dot is what makes the address address, whether it is a "valid" address worth highlighting (because it points to a block or attribute which actually exists) depends on the individual segments, although the dot is important pre-requisite. A more practical reason is that users will likely want the dots to be highlighted differently from the segments.

decoder/semantic_tokens_expr_test.go
```diff
@@ -1,16 +1,18 @@
 package decoder
 
 import (
+	"context"
 	"fmt"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/zclconf/go-cty/cty"
+
```
radeksimko, line 10:
```suggestion
```

decoder/semantic_tokens_expr_test.go
```diff
@@ -1433,6 +1435,8 @@ EOT
 		},
 	}
 
+	ctx := context.Background()
```
radeksimko, line 1438: Although it won't make a difference in this particular test, it's typically best to start each (sub)test with a fresh/isolated context, i.e. move this under `t.Run()` below.

decoder/semantic_tokens_test.go
```diff
@@ -1,16 +1,18 @@
 package decoder
 
 import (
+	"context"
 	"errors"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/zclconf/go-cty/cty"
+
```
radeksimko, line 10:
```suggestion
```

decoder/semantic_tokens_test.go
```diff
@@ -836,3 +852,204 @@ resource "vault_auth_backend" "blah" {
 		t.Fatalf("unexpected tokens: %s", diff)
 	}
 }
+
+func TestDecoder_SemanticTokensInFile_expression_extensions(t *testing.T) {
```
radeksimko, line 856:
```suggestion
func TestDecoder_SemanticTokensInFile_extensions(t *testing.T) {
```

decoder/semantic_tokens_test.go
```diff
@@ -836,3 +852,204 @@ resource "vault_auth_backend" "blah" {
...
+					},
+					{Name: "name"},
+				},
+			},
+		},
+	}
+
+	testCfg := []byte(`
+resource "vault_auth_backend" "blah" {
+	count = 1
```
radeksimko, line 886: This looks a bit misaligned - probably a mix of tabs and spaces?

decoder/semantic_tokens_test.go
```diff
@@ -836,3 +852,204 @@ resource "vault_auth_backend" "blah" {
...
+					Byte:   53,
+				},
+				End: hcl.Pos{
+					Line:   4,
+					Column: 12,
+					Byte:   62,
+				},
+			},
+		},
+		{ // count.
```
radeksimko, line 1015:
```suggestion
		{ // count
```

decoder/semantic_tokens_test.go
```diff
@@ -836,3 +852,204 @@ resource "vault_auth_backend" "blah" {
...
+					Byte:   65,
+				},
+				End: hcl.Pos{
+					Line:   4,
+					Column: 20,
+					Byte:   70,
+				},
+			},
+		},
+		{ // .index
```
radeksimko, line 1032: The comment reflects the current implementation correctly but as mentioned, we typically only report the segments as segments, not the separator.

schema/body_schema.go
```diff
@@ -45,6 +45,29 @@ type BodySchema struct {
...
+
+	// Extensions represents any HCL extensions supported in this body
+	Extensions *BodyExtensions
+}
+
+type BodyExtensions struct {
+	Count         bool // count attribute + count.index refs
+	ForEach       bool // for_each attribute + each.* refs
+	DynamicBlocks bool // dynamic "block-name" w/ content & for_each inside
+	SelfRefs      bool // self.* refs which refer to outermost parent block body
```
radeksimko, line 57:
```suggestion
```
Assuming this PR focuses on just `count`, so we can leave out the other extensions for now?

radeksimko (commented):
Looks pretty good and functional - just needs some cleanup - I left some comments in-line.

decoder/body_candidates.go
```diff
@@ -135,3 +146,17 @@ func isBlockDeclarable(body *hclsyntax.Body, blockType string, bSchema *schema.B
 	}
 	return true
 }
+
+func countAttributeCandidate(editRng hcl.Range) lang.Candidate {
+	return lang.Candidate{
+		Label:       "count",
+		Detail:      "optional, number",
+		Description: lang.PlainText("The distinct index number (starting with 0) corresponding to the instance"),
```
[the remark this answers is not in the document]
jpogran: Where would you like the extracted construct placed?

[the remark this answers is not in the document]
radeksimko: An un-exported variable within `decoder` may be fine for now? I guess `schema` package would be even better, but we can always revisit as part of implementing the other extensions.

decoder/semantic_tokens.go
jpogran: Addressed!

decoder/expression_candidates.go
```diff
@@ -457,9 +459,24 @@ func (d *PathDecoder) constraintToCandidates(constraint schema.ExprConstraint, o
 	return candidates
 }
 
-func (d *PathDecoder) candidatesForTraversalConstraint(tc schema.TraversalExpr, outerBodyRng, prefixRng, editRng hcl.Range) []lang.Candidate {
+func (d *PathDecoder) candidatesForTraversalConstraint(ctx context.Context, tc schema.TraversalExpr, outerBodyRng, prefixRng, editRng hcl.Range) []lang.Candidate {
 	candidates := make([]lang.Candidate, 0)
 
+	ext, ok := icontext.ExtensionsFromContext(ctx)
```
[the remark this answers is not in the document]
jpogran: Removed

context/context.go
```diff
@@ -0,0 +1,28 @@
+package context
```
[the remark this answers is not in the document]
jpogran: While we can move these to the schema package, I can't hide them because they're used in decoder almost exclusively. Did you mean to move them there?

decoder/semantic_tokens_test.go
```diff
@@ -836,3 +851,556 @@ resource "vault_auth_backend" "blah" {
...
+				},
+			},
+		},
+	}
+
+	diff := cmp.Diff(expectedTokens, tokens)
+	if diff != "" {
+		t.Fatalf("unexpected tokens: %s", diff)
+	}
+}
```
radeksimko, line 1406: Do you mind adding a test for `count.index` in a sub-block? e.g.

```
resource "foobar" "name" {
  count = 1
  block {
    attr = count.index
  }
}
```
I believe it should pass with the current implementation, but it would be nice to verify the context propagation.

context/context.go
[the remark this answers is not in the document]
radeksimko: I guess one way to address that would be hiding under `schema/internal` but that introduces potential new problems with import cycles, so I'd just leave it as-is.

dbanck (approved):
Latest test LGTM!

PR #139, opened 2022-10-17
Support Dynamic block expression extensions

jpogran:
Adds support for completion of dynamic blocks in resource, data, provider and provisioner blocks.

This does not provide completion for the label for a given dynamic block, which will be provided in a future set of work.

Partially implements https://github.com/hashicorp/terraform-ls/issues/530

### DIFF
--- a/decoder/decoder.go
+++ b/decoder/decoder.go
@@ -238,6 +238,7 @@ func dynamicBlockSchema() *schema.BlockSchema {
 			Blocks: map[string]*schema.BlockSchema{
 				"content": {
 					Description: lang.PlainText("The body of each generated block"),
+					MaxItems:    1,
 				},
 			},
 		},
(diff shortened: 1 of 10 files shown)


decoder/decoder.go
```diff
@@ -195,6 +195,55 @@ func forEachAttributeSchema() *schema.AttributeSchema {
...
+						},
+					},
+					IsOptional: true,
+					Description: lang.Markdown("A list of strings that specifies the block labels, " +
+						"in order, to use for each generated block."),
+				},
+			},
+			Blocks: map[string]*schema.BlockSchema{
+				"content": {
+					Description: lang.PlainText("The body of each generated block"),
```
dbanck, line 240:
```suggestion
					Description: lang.PlainText("The body of each generated block"),
					MaxItems:    1,
```

We can set `MaxItems` here, so autocompletion only suggests a single `content` block.

dbanck (approved):
LGTM 👍 Just a small suggestion

PR #142, opened 2022-10-19
foreach extensions

jpogran:
- wip
- decoder: add failing tests for extensions w/ DependentBody
- fix merging schema extensions

### DIFF
--- a/decoder/body_extensions_test.go
+++ b/decoder/body_extensions_test.go
@@ -365,6 +365,202 @@ func TestCompletionAtPos_BodySchema_Extensions(t *testing.T) {
 				},
 			}),
 		},
+		{
+			"count attribute completion with DependentBody",
+			&schema.BodySchema{
+				Blocks: map[string]*schema.BlockSchema{
+					"resource": {
+						Labels: []*schema.LabelSchema{
+							{
+								Name:     "type",
+								IsDepKey: true,
+							},
+							{
+								Name: "name",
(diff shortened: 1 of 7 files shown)


decoder/body_extensions_test.go
```diff
@@ -533,7 +536,7 @@ func TestCompletionAtPos_BodySchema_Extensions(t *testing.T) {
 							},
 						},
 						NewText: "for_each",
-						Snippet: "for_each = ${1:1}",
+						Snippet: "for_each {\n ${1}\n}",
```
radeksimko, line 539: I think the original snippet was correct - `for_each` is not a block, but an attribute, so there's no need for the curly braces and newlines.

jpogran: I was following the first example in https://www.terraform.io/language/meta-arguments/for_each#basic-syntax. I missed the quals sign, but thought since it shows map first that is the likely first choice to use

jpogran:
This had too many conflicts with https://github.com/hashicorp/hcl-lang/pull/144, so I've closed this in favor of a fresh start in https://github.com/hashicorp/hcl-lang/pull/145

jpogran:
This had too many conflicts with https://github.com/hashicorp/hcl-lang/pull/144, so I've closed this in favor of a fresh start in https://github.com/hashicorp/hcl-lang/pull/145

PR #144, opened 2022-10-21
decoder: fix completion & hover for `count` extension value

radeksimko:
I'd love to attach some screenshots before & after, but all usage of the extensions currently involves `DependentBody` which itself has some bugs, which are being fixed as part of https://github.com/hashicorp/hcl-lang/pull/142/commits/32e493e09ce61eca33e4fa2176e943ac599a9015 in https://github.com/hashicorp/hcl-lang/pull/142

So failing & passing tests will hopefully suffice.

### DIFF
--- a/decoder/body_extensions_test.go
+++ b/decoder/body_extensions_test.go
@@ -6,12 +6,13 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/zclconf/go-cty/cty"
+
 	"github.com/hashicorp/hcl-lang/lang"
 	"github.com/hashicorp/hcl-lang/reference"
 	"github.com/hashicorp/hcl-lang/schema"
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclsyntax"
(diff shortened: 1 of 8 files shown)


decoder/body_extensions_test.go
```diff
@@ -6,12 +6,12 @@ import (
...
 
 	"github.com/google/go-cmp/cmp"
-	"github.com/zclconf/go-cty/cty"
-
 	"github.com/hashicorp/hcl-lang/lang"
+	"github.com/hashicorp/hcl-lang/reference"
 	"github.com/hashicorp/hcl-lang/schema"
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclsyntax"
+	"github.com/zclconf/go-cty/cty"
```
jpogran, line 14: I thought we didn't want to move these?

jpogran (approved):
Works as advertised!

decoder/body_extensions_test.go
radeksimko: Good point.

PR #149, opened 2022-11-04
Support for local references

radeksimko:
## TODO

Reference Matching Logic:

 - [x] fix panic when trying to shorten a local reference - @radeksimko 
 - [x] fix mismatch of global nested target with local address set - @radeksimko 

Completion:

 - [x] fix broken `count.index` completion - @dbanck 

Self References:
 - [x] Filter invalid self during origin collection
 - [x] Fix broken references output
 - [x] Use localAddr for hover
 - [x] Use localAddr for autocompletion
 - [x] Use localAddr for semantic tokens
 - [x] Refactoring to use collected origins
   - [x] for hover
   - [x] for semantic tokens
 - [ ] Add more tests

### DIFF
--- a/decoder/body_extensions_test.go
+++ b/decoder/body_extensions_test.go
@@ -60,7 +60,7 @@ func TestCompletionAtPos_BodySchema_Extensions(t *testing.T) {
 				{
 					Label: "count",
 					Description: lang.MarkupContent{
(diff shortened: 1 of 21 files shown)


radeksimko:
Closing in favour of https://github.com/hashicorp/hcl-lang/pull/163 and other PRs which were already merged in.

PR #150, opened 2022-11-04
Add support for local reference targets

radeksimko:
This PR intended to be no-op for the end-user.

It introduces the concept of "local references", which can be utilised for `count.index`, `each.*` and `self.*` references, as drafted within https://github.com/hashicorp/hcl-lang/pull/149

While there are tests attached, there is no code actually collecting such local targets - that is left for follow-up PRs which can be made out of the linked PR.

It also leaves out the question of rendering such reference targets. Now that there can be two ways of referring to the same target, there can be some contexts where it is appropriate to use one address over the other, and vice versa. I could not find any way to solve that problem in an abstract sense which would work for all three use cases. So I left this for the follow-up PRs. I did however leave TODOs in a few places where it is likely this problem will need to be addressed + I introduced an `Address()` method to `reference.Target`, which should essentially "just work" for the simpler cases of `count.index` and `each.*`.

### DIFF
--- a/decoder/expression_candidates.go
+++ b/decoder/expression_candidates.go
@@ -478,22 +478,23 @@ func (d *PathDecoder) candidatesForTraversalConstraint(tc schema.TraversalExpr, 
 
 	prefix, _ := d.bytesFromRange(prefixRng)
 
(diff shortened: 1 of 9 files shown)


dbanck (approved):
LGTM 👍

PR #155, opened 2022-11-15
Refactor `hoverAtPos` to use collected origins

dbanck:
This change ensures that we only use the collected origins when matching targets on hover. So whenever we decide not to collect an origin during collection, the hover logic will also exclude it.


radeksimko (approved):
LGTM, thanks

PR #156, opened 2022-11-15
Refactor `semanticTokensInFile` to use collected origins

dbanck:
This change ensures that we only use the collected origins when building semantic tokens. So whenever we decide not to collect an origin during collection, the semantic token logic will also exclude it.


radeksimko (approved):
LGTM, one thing, which I also forgot to mention in your previous PR for hover is that we should probably add a negative test which would otherwise fail without this new logic - i.e. confirm that no hover and no semantic highlighting happens when there's no origin, but a possible matching target is available.

dbanck:
Good catch! I've added two test cases to account for that.

PR #157, opened 2022-11-15
reference: Fix matching for targets which involve ranges & both addresses

radeksimko:
As the test shows, we'd previously consider `TargetableFromRangePtr` mismatch as complete mismatch, even though the same target could be a match using the absolute address.

Mostly for convenience in test, the implementation treats `TargetableFromRangePtr: nil` as an automatic match, but I can be convinced to change that, if we deem it's a bad idea.

### DIFF
--- a/reference/target.go
+++ b/reference/target.go
@@ -187,10 +187,12 @@ func (target Target) Matches(origin MatchableOrigin) bool {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


dbanck (approved):
LGTM 👍 

I think it's fine to treat `TargetableFromRangePtr == nil` as an automatic match.

PR #158, opened 2022-11-15
reference: Fix panic when trying to shorten origin address

radeksimko:
As the in-line comment explains, previously we'd try to "shorten" the origin address by actually "growing" it.

In theory this could have happened prior to https://github.com/hashicorp/hcl-lang/pull/150 but it apparently never did, or nobody reported it. It would require an origin with less segments than the compared target. In most (all?) cases, the targets have at least 2 segments and so that would require comparison with a single-segment origin, such as `foo`. This was also additionally gated by another condition, which was somewhat naively modified to account for local addresses and that condition is no longer necessary now.

### DIFF
--- a/reference/target.go
+++ b/reference/target.go
@@ -150,10 +150,6 @@ func (ref Target) ConformsToType(typ cty.Type) bool {
 }
 
 func (target Target) Matches(origin MatchableOrigin) bool {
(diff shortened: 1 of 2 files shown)


dbanck (approved):
LGTM 👍

PR #159, opened 2022-11-21
reference: refactor & fix matching of cyclical refs

radeksimko:
This introduces some minor changes to the interfaces, to plumb context through, in preparation for using it to pass down context, such as whether "SelfRefs" is active, i.e. whether `self.*` addresses are relevant in that context.

Additionally, the matching logic is being refactored and updated to account for cyclical references, specifically to continue **rejecting** completion such as

```hcl
resource "aws_instance" "test" {
  count = 4

  cpu_core_count = aws_instance.test.cpu_core_count
}
```
**TODO**: Address hover & semtok highlighting which still _recognize_ the above as valid ^

and

```hcl
resource "aws_instance" "test" {
  count = count.index
}
```

and (future)

```hcl
resource "aws_instance" "test" {
  ami = self.ami
}
```


dbanck (approved):
Nice! Thanks for extracting this

PR #160, opened 2022-11-21
Use local ref. targets for `count.index`

radeksimko:
Depends on #159 

--- 

This PR changes collection of reference targets, such that we now collect `count.index` as a target to the `count` attribute. This in turn enables go-to-definition and go-to-references between `count` and `count.index`.

Additionally, it updates completion, hover and semantic tokens, to make use of this reference target and just treat it like any other valid reference.

Lastly I updated the description of the `count` attribute to reflect what it does and also add the note about conflict we already have in the `for_each` description.

--- 

### Completion (w/ updated description)

Screenshot 2022-11-21 at 17 52 19
Screenshot 2022-11-21 at 18 00 21

### Hover (w/ updated description)

Screenshot 2022-11-21 at 17 52 45
Screenshot 2022-11-21 at 17 57 21

### DIFF
--- a/decoder/body_extensions_test.go
+++ b/decoder/body_extensions_test.go
@@ -60,7 +60,7 @@ func TestCompletionAtPos_BodySchema_Extensions(t *testing.T) {
 				{
 					Label: "count",
 					Description: lang.MarkupContent{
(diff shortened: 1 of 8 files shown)


dbanck (approved):
LGTM! Great job on outlining the UX impact

PR #161, opened 2022-11-22
Avoid reporting cyclical references in hover/semtok/go-to-*

radeksimko:
Currently, the reference in the configuration below is treated as valid in the context of hover, semantic token highlighting and go-to-definition/go-to-references, even though it's not a valid reference.

```hcl
resource "aws_instance" "test" {
  ami = aws_instance.test.ami
}
```

### DIFF
--- a/decoder/reference_origins_collect_hcl_test.go
+++ b/decoder/reference_origins_collect_hcl_test.go
@@ -1015,6 +1015,125 @@ tup = [ var.three ]
 				},
 			},
 		},
+		{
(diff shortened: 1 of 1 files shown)


radeksimko:
There is some logic handling address resolution which we could share between target and origin collection, but it would need to be refactored (out).

I think ideally, we should end up with 1 or more functions which walk the AST (`hcl.Body`) & schema, and keep track of the address in any given context - in case of targets we'd just collect the address and append it to the slice, in case of origin we'd compare it with the traversal for equality.

PR #162, opened 2022-11-22
Use local ref. targets for `each.*` references

radeksimko:
This PR changes collection of reference targets, such that we now collect `each.key` and `each.value` as targets to the `for_each` attribute. This in turn enables go-to-definition and go-to-references between `for_each` and the two references.

Additionally, it updates completion, hover and semantic tokens, to make use of these reference targets and just treat them like any other valid reference.

--- 

### Completion (same as before)

Screenshot 2022-11-22 at 13 34 48
Screenshot 2022-11-22 at 13 35 05

### Hover (same as before)

Screenshot 2022-11-22 at 13 37 35
Screenshot 2022-11-22 at 13 37 12

### DIFF
--- a/decoder/body_extensions_test.go
+++ b/decoder/body_extensions_test.go
@@ -693,7 +693,44 @@ variable "test" {
 					},
 				},
 			},
-			reference.Targets{},
+			reference.Targets{
+				{
+					LocalAddr: lang.Address{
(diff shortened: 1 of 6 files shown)


dbanck (approved):
LGTM 👍

PR #165, opened 2022-11-25
Avoid completing static block inside a `dynamic` label

dbanck:
This PR fixes an error where we're offering static blocks for completion when there are no blocks in the dependent body.

## Before
Screenshot 2022-11-24 at 15 59 21

## After
CleanShot 2022-11-29 at 17 39 00
`dynamic` isn't available anymore, because there are no blocks in the dependent body.

### DIFF
--- a/decoder/decoder.go
+++ b/decoder/decoder.go
@@ -50,10 +50,6 @@ func mergeBlockBodySchemas(block *hcl.Block, blockSchema *schema.BlockSchema) (*
 		mergedSchema.ImpliedOrigins = make([]schema.ImpliedOrigin, 0)
 	}
 
-	if mergedSchema.Extensions != nil && mergedSchema.Extensions.DynamicBlocks && len(mergedSchema.Blocks) > 0 {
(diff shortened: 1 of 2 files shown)


decoder/decoder.go
```diff
@@ -97,6 +93,10 @@ func mergeBlockBodySchemas(block *hcl.Block, blockSchema *schema.BlockSchema) (*
 		if depSchema.Extensions != nil {
 			mergedSchema.Extensions = depSchema.Extensions.Copy()
 		}
+	} else {
+		if mergedSchema.Extensions != nil && mergedSchema.Extensions.DynamicBlocks && len(mergedSchema.Blocks) > 0 {
+			mergedSchema.Blocks["dynamic"] = buildDynamicBlockSchema(mergedSchema)
+		}
```
radeksimko, line 99: Could this be turned into `else if`, just to aid the readability a little bit?

radeksimko (approved):
LGTM, just one minor suggestion.

PR #166, opened 2022-11-28
Fix missing hover for `count` and `for_each` expression

dbanck:
This PR adds the extension-based schema changes for `count` and `for_each` during reference origin collection. This error occurred because we collected no origins for `var.name`.

## Before
CleanShot 2022-11-28 at 16 51 30@2x

CleanShot 2022-11-28 at 16 51 02@2x

## After
CleanShot 2022-11-30 at 11 14 48@2x

CleanShot 2022-11-30 at 11 15 31@2x

### DIFF
--- a/decoder/hover_test.go
+++ b/decoder/hover_test.go
@@ -935,7 +935,7 @@ func TestDecoder_HoverAtPos_typeDeclaration(t *testing.T) {
 	}
 }
 
-func TestDecoder_HoverAtPos_extension(t *testing.T) {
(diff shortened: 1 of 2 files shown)


radeksimko (approved):
I'm guessing this was lost in the large `dynamic` PR somewhere, when we have made the origin collection stricter, right?

Either way, nice fix! 👍🏻 🚢

dbanck:
Yeah, the changes introduced in https://github.com/hashicorp/hcl-lang/pull/155 require a valid collected origin for hover now. And we never created one for `count` and `for_each`.

PR #167, opened 2022-11-30
fix: support of deeper nesting of dynamic block

radeksimko:
### Completion
Screenshot 2022-11-30 at 11 59 46

### Hover
Screenshot 2022-11-30 at 12 00 52

### Semantic Tokens
Screenshot 2022-11-30 at 12 01 13

### Semantic Tokens
Screenshot 2022-11-30 at 12 04 58

### DIFF
--- a/decoder/body_extensions_test.go
+++ b/decoder/body_extensions_test.go
@@ -2553,7 +2553,6 @@ resource "aws_elastic_beanstalk_environment" "example" {
 			}),
 			"",
 		},
-		// never complete dynamic as a dynamic label
 		{
 			"never complete dynamic as a dynamic label",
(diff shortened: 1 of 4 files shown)


decoder/body_extensions_test.go
```diff
@@ -2553,7 +2553,6 @@ resource "aws_elastic_beanstalk_environment" "example" {
 			}),
 			"",
 		},
-		// never complete dynamic as a dynamic label
```
radeksimko, line 2556: The reason I removed this comment here is because it doesn't seem to be adding any value, as it just duplicates the name of the test as below.

dbanck (approved):
Nice fix!

PR #174, opened 2022-12-08
decoder/schema: decouple extension-related logic

radeksimko:
This is just a minor cleanup, prior to making some other, potentially bigger changes in expressions.

### DIFF
--- a/schema/context.go
+++ b/schema/context.go
@@ -4,18 +4,6 @@ import (
 	"context"
 )
 
-type bodyExtCtxKey struct{}
-
-type bodyActiveCountCtxKey struct{}
-
-func WithActiveCount(ctx context.Context) context.Context {
-	return context.WithValue(ctx, bodyActiveCountCtxKey{}, true)
-}
-
-func ActiveCountFromContext(ctx context.Context) bool {
-	return ctx.Value(bodyActiveCountCtxKey{}) != nil
-}
-
 type bodyActiveForEachCtxKey struct{}
 
 func WithActiveForEach(ctx context.Context) context.Context {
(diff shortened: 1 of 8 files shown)


schema/context.go
```diff
@@ -0,0 +1,37 @@
...
+
+type bodyExtCtxKey struct{}
+
+type bodyActiveCountCtxKey struct{}
+
+func WithActiveCount(ctx context.Context) context.Context {
+	return context.WithValue(ctx, bodyActiveCountCtxKey{}, true)
+}
+
+func ActiveCountFromContext(ctx context.Context) bool {
```
dbanck, line 15: This function isn't used anywhere. Either we missed adding it somewhere or the refactoring made most of the context related checks obsolete.

schema/context.go
```diff
@@ -0,0 +1,37 @@
...
+	return ctx.Value(bodyActiveCountCtxKey{}) != nil
+}
+
+type bodyActiveForEachCtxKey struct{}
+
+func WithActiveForEach(ctx context.Context) context.Context {
+	return context.WithValue(ctx, bodyActiveForEachCtxKey{}, true)
+}
+
+func ActiveForEachFromContext(ctx context.Context) bool {
```
dbanck, line 25: This function is only used within the semantic tokens:
https://github.com/hashicorp/hcl-lang/blob/6bf6bd83e2bc2b61aa8df2aab8ce4e60c0776852/decoder/semantic_tokens.go#L176-L193

But with the matching of references and origins, it might not be required anymore to detect a valid traversal step.

dbanck (commented):
We might be able to remove some of the context related functions. What do you think?

PR #175, opened 2022-12-08
Remove `schema.TupleConsExpr`

radeksimko:
Related `terraform-schema` PR: https://github.com/hashicorp/terraform-schema/pull/169

--- 

Originally the `TupleConsExpr` constraint was meaning to express the "raw" HCL `TupleConsExpr` (essentially the `[]` square brackets) in context where the expression isn't interpreted as data.

However, as we will be making some bigger changes to expressions, this additional constraint is something we'd have to account for too, when we can instead just use any of the existing types, such as `ListExpr`, `SetExpr` or `TupleExpr`, on best-effort basis to express how the expression is treated and reduce the total number of constraints we have to deal with.

For example, in Terraform's `data` or `resource` block, we have `depends_on = []`. We can choose `SetExpr` with the reasoning that ordering of elements within doesn't matter and every element is of the same type.

https://github.com/hashicorp/terraform-schema/pull/169 also uncovered one minor bug with rendering of collections types with multiple element constraints, so I added some more tests and fixed the bug.


dbanck (approved):
LGTM 👍

PR #182, opened 2023-01-16
Refactor `NewExpression` to allow access to path context

dbanck:
This is a small update to #177 and enables access to path context, which is required for working with functions, references and origins.

### DIFF
--- a/schema/constraint_any_expression.go
+++ b/schema/constraint_any_expression.go
@@ -4,9 +4,13 @@ import (
 	"github.com/zclconf/go-cty/cty"
 )
 
-// AnyExpression TODO
+// AnyExpression represents any expression type convertible
+// to the given data type (OfType).
+//
+// For example function call returning cty.String complies with
+// AnyExpression{OfType: cty.String}.
 type AnyExpression struct {
-	// OfType defines the type of a type-aware reference
+	// OfType defines the type which the outermost expression is constrained to
 	OfType cty.Type
 }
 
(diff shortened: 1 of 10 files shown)


schema/constraint_any_expression.go
```diff
@@ -0,0 +1,30 @@
+package schema
+
+import (
+	"github.com/zclconf/go-cty/cty"
+)
+
+// AnyExpression TODO
```
radeksimko, line 7:
```suggestion
// AnyExpression represents any expression type convertible
// to the given data type (OfType).
//
// For example function call returning cty.String complies with
// AnyExpression{OfType: cty.String}.
```

schema/constraint_any_expression.go
```diff
@@ -0,0 +1,30 @@
...
+type AnyExpression struct {
+	// OfType defines the type of a type-aware reference
```
radeksimko, line 9:
```suggestion
	// OfType defines the type which the outermost expression is constrained to
```

schema/constraint_any_expression.go
```diff
@@ -0,0 +1,30 @@
+package schema
+
+import (
+	"github.com/zclconf/go-cty/cty"
+)
+
+// AnyExpression TODO
+type AnyExpression struct {
```
radeksimko, line 8: Can you also add a basic compilation test to `schema/constraint_test.go` for this new constraint?

e.g.
```go
_ Constraint = AnyExpression{}
```
just to make sure it complies with the `Constraint` interfaces and continues to be compliant.

radeksimko (approved):
LGTM aside from the few minor in-line comments.

radeksimko (approved):

PR #185, opened 2023-01-16
decoder: Add support for `Reference` as `Constraint`

radeksimko:

--- 

There is quite a lot of logic related to reference collection, which I am not touching in this PR. That will be addressed more broadly as part of https://github.com/hashicorp/hcl-lang/pull/239 I did however go through all the related legacy tests and made sure they pass (when using `schema.Reference`) in combination with this PR.

The updated tests are in https://github.com/hashicorp/hcl-lang/commit/c0a2da2fe7f45573798860481d1f62bb363e4407

---

### DIFF
--- a/decoder/expr_reference.go
+++ b/decoder/expr_reference.go
@@ -15,11 +15,6 @@ type Reference struct {
 	pathCtx *PathContext
 }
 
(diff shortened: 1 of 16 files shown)


jpogran (approved):
🚢

PR #186, opened 2023-01-16
decoder: Add support for `LiteralType` as `Constraint`

radeksimko:

--- 

Existing tests were updated as part of https://github.com/hashicorp/hcl-lang/commit/699069384a9c1ef08b7b7a7ab9e8cd6eb8ac739b (not part of this PR) and yield either comparable or better results, except for reference collection, which:

 - we agreed to leave for a separate PR
 - involves other constraints which aren't implemented yet

---

### DIFF
--- a/decoder/expr_literal_type_hover.go
+++ b/decoder/expr_literal_type_hover.go
@@ -1,0 +1,149 @@
+package decoder
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/hcl-lang/lang"
+	"github.com/hashicorp/hcl-lang/schema"
+	"github.com/hashicorp/hcl/v2"
+	"github.com/hashicorp/hcl/v2/hclsyntax"
+	"github.com/zclconf/go-cty/cty"
+)
+
+func (lt LiteralType) HoverAtPos(ctx context.Context, pos hcl.Pos) *lang.HoverData {
+	typ := lt.cons.Type
+
+	if typ == cty.DynamicPseudoType {
+		val, _ := lt.expr.Value(nil)
+		if val.IsKnown() && !val.IsNull() {
+			typ = val.Type()
+		}
+	}
+
+	// string is a special case as it's always represented like a template
+	// even if there's no templating involved
+	if typ == cty.String {
+		expr, ok := lt.expr.(*hclsyntax.TemplateExpr)
+		if !ok {
+			return nil
+		}
+		if expr.IsStringLiteral() || isMultilineStringLiteral(expr) {
+			return &lang.HoverData{
+				Content: lang.Markdown(fmt.Sprintf(`_%s_`, typ.FriendlyName())),
+				Range:   expr.Range(),
+			}
+		}
+
+		return nil
+	}
+
+	if typ.IsPrimitiveType() {
+		expr, ok := lt.expr.(*hclsyntax.LiteralValueExpr)
+		if !ok {
+			return nil
+		}
+
+		if !lt.cons.Type.Equals(expr.Val.Type()) {
+			return nil
+		}
+
+		return &lang.HoverData{
+			Content: lang.Markdown(fmt.Sprintf(`_%s_`, typ.FriendlyName())),
+			Range:   expr.Range(),
+		}
+	}
+
+	if typ.IsListType() {
+		expr, ok := lt.expr.(*hclsyntax.TupleConsExpr)
+		if !ok {
+			return nil
+		}
+
+		cons := schema.List{
+			Elem: schema.LiteralType{
+				Type: typ.ElementType(),
+			},
+		}
+
+		return newExpression(lt.pathCtx, expr, cons).HoverAtPos(ctx, pos)
+	}
+
+	if typ.IsSetType() {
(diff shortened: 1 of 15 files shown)


decoder/expr_literal_type_hover.go
```diff
@@ -0,0 +1,149 @@
...
+)
+
+func (lt LiteralType) HoverAtPos(ctx context.Context, pos hcl.Pos) *lang.HoverData {
+	typ := lt.cons.Type
+
+	if typ == cty.DynamicPseudoType {
+		val, _ := lt.expr.Value(nil)
+		if val.IsKnown() && !val.IsNull() {
+			typ = val.Type()
+		}
```
radeksimko, line 21:
```suggestion
		val, diags := lt.expr.Value(nil)
		if !diags.HasErrors() {
			typ = val.Type()
		}
```
Should be more or less equivalent, but this is probably more correct.

decoder/expr_literal_type_semtok.go
```diff
@@ -0,0 +1,153 @@
...
+)
+
+func (lt LiteralType) SemanticTokens(ctx context.Context) []lang.SemanticToken {
+	typ := lt.cons.Type
+
+	if typ == cty.DynamicPseudoType {
+		val, _ := lt.expr.Value(nil)
+		if val.IsKnown() && !val.IsNull() {
+			typ = val.Type()
+		}
```
radeksimko, line 19:
```suggestion
		val, diags := lt.expr.Value(nil)
		if !diags.HasErrors() {
			typ = val.Type()
		}
```

decoder/expr_literal_type_semtok.go
```diff
@@ -0,0 +1,153 @@
...
+			return []lang.SemanticToken{
+				{
+					Type:      lang.TokenString,
+					Modifiers: lang.SemanticTokenModifiers{},
+					Range:     expr.Range(),
+				},
+			}
+		}
+
+		// TODO: report multiline/HEREDOC notation
```
jpogran, line 39: Should this be ticketed?

decoder/expr_literal_type_hover.go
```diff
@@ -0,0 +1,149 @@
...
+
+	"github.com/hashicorp/hcl-lang/lang"
+	"github.com/hashicorp/hcl-lang/schema"
+	"github.com/hashicorp/hcl/v2"
+	"github.com/hashicorp/hcl/v2/hclsyntax"
+	"github.com/zclconf/go-cty/cty"
+)
+
+func (lt LiteralType) HoverAtPos(ctx context.Context, pos hcl.Pos) *lang.HoverData {
+	typ := lt.cons.Type
```
jpogran, line 15: Is there a case where cons can be nil? Just to guard against a panic here

decoder/expr_literal_type_ref_targets.go
```diff
@@ -0,0 +1,114 @@
...
+import (
+	"context"
+
+	"github.com/hashicorp/hcl-lang/reference"
+	"github.com/hashicorp/hcl-lang/schema"
+	"github.com/hashicorp/hcl/v2/hclsyntax"
+)
+
+func (lt LiteralType) ReferenceTargets(ctx context.Context, targetCtx *TargetContext) reference.Targets {
+	typ := lt.cons.Type
```
jpogran, line 12: Is there a case where cons can be nil? Just to guard against a panic here

decoder/expr_literal_type_semtok.go
```diff
@@ -0,0 +1,153 @@
...
+	"context"
+
+	"github.com/hashicorp/hcl-lang/lang"
+	"github.com/hashicorp/hcl-lang/schema"
+	"github.com/hashicorp/hcl/v2/hclsyntax"
+	"github.com/zclconf/go-cty/cty"
+)
+
+func (lt LiteralType) SemanticTokens(ctx context.Context) []lang.SemanticToken {
+	typ := lt.cons.Type
```
jpogran, line 13: Is there a case where cons can be nil? Just to guard against a panic here

jpogran (approved):
Small concerns. Great work!

decoder/expr_literal_type_ref_targets.go
radeksimko: It's a good point, although that would indicate invalid schema. So I think the best place to cover that would be in the schema validation. I will add some checks there!

decoder/expr_literal_type_semtok.go
radeksimko: It's a good point, although that would indicate invalid schema. So I think the best place to cover that would be in the schema validation. I will add some checks there!

radeksimko:
I added `LiteralType` validation for nil `Type` and also another one to ensure there's either `Constraint` or `Expr`, so that should ensure integrity on a higher level and ensure there's no `nil` `Constraint`.

If we still somehow end up with `nil` constraint further down then I would call that a really bad bug which needs to be fixed in the place that produces the schema, not on our side. In other words, we shouldn't be producing invalid schema in the first place anywhere and then hiding that fact, so crash is the right thing to do in such a case.

PR #187, opened 2023-01-16
decoder: Add support for `LiteralValue` as `Constraint`

radeksimko:

---


radeksimko (commented):
LGTM :shipit: 

Screenshot 2023-03-30 at 18 03 53

PR #188, opened 2023-01-16
decoder: Extract & document `isEmptyExpression()`

radeksimko:
This helper is expected to be used in completion for most constraints.

I am still not fully confident it covers all cases - as explained in the comment, but I do believe it deserves to be extracted.


dbanck (approved):
LGTM! 

I never got to find out if more checks are necessary for detecting an empty expression.

PR #189, opened 2023-01-16
decoder: Refactor `newExpression()` & plumb `PathContext`

radeksimko:
This is primarily to make it easier to "hand-over" completion/hover/etc. between different expression types (which do not /need to/ have access to all of `PathDecoder`) and reduce diff conflicts between upcoming PRs where each involves different expression type.

It is expected that all complex types (list, set, tuple, map, object) _may_ contain either `Reference` (hence need `ReferenceTargets`) or `AnyExpression` (hence need future `FunctionsMap` and possibly `ReferenceTargets` too).

I have briefly considered creating a separate type, e.g. `ExpressionContext` or pass these things via `context.Context`, or use the static `newExpression()` in all places where `(*PathDecoder) newExpression()` is now used,  but concluded that it would all be premature optimization at this point.

### DIFF
--- a/decoder/expr_list.go
+++ b/decoder/expr_list.go
@@ -10,8 +10,9 @@ import (
 )
 
 type List struct {
-	expr hcl.Expression
-	cons schema.List
+	expr    hcl.Expression
+	cons    schema.List
+	pathCtx *PathContext
 }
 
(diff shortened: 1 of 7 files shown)


dbanck (approved):
👍 LGTM

PR #190, opened 2023-01-18
decoder: Introduce `newEmptyExpressionAtPos()`

radeksimko:
As per comment:

`newEmptyExpressionAtPos` returns a new "artificial" empty expression which can be used during completion inside of another expression in an empty space which isn't already represented by empty expression.

For example, new argument after comma in function call, or new element in a list or set.


dbanck (approved):
👍

PR #191, opened 2023-01-23
decoder: Make reference collection optional for Expression

radeksimko:
This better aligns with a convention of minimal interfaces in Go, also surfaced as a [Go Proverb](https://go-proverbs.github.io/) ["The bigger the interface, the weaker the abstraction."](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=5m17s) and exemplified throughout the stdlib, e.g. in the [`io/fs.FS`](https://pkg.go.dev/io/fs#FS) interface and [friends](https://pkg.go.dev/io/fs#ReadDirFS).

Another (AFAIK unwritten) convention is to suffix such interface types with `-er` (e.g. `io.Writer`, `io.Reader`), but as shown on the `fs` interfaces, this isn't always practical and IMO wouldn't be in our case.

To my best knowledge all types we have bootstrapped currently _would_ still need to provide completion, hover and semantic tokens, so I'm keeping these in the base interface.

### DIFF
--- a/decoder/expr_keyword.go
+++ b/decoder/expr_keyword.go
@@ -4,7 +4,6 @@ import (
 	"context"
 
 	"github.com/hashicorp/hcl-lang/lang"
-	"github.com/hashicorp/hcl-lang/reference"
 	"github.com/hashicorp/hcl-lang/schema"
 	"github.com/hashicorp/hcl/v2"
 )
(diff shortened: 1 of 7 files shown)


dbanck (approved):
LGTM 👍

PR #192, opened 2023-01-23
schema: Make Elem(s) optional for list/set/map constraint

radeksimko:
I noticed while testing https://github.com/hashicorp/hcl-lang/pull/184 that the current implementation of `List`, `Set` and `Map` constraints does not allow for elements to be undefined - i.e. it would currently cause a crash.

We could theoretically enforce non-empty `Elem`s via static validation, but in practice we actually have a limited use case - `ignore_changes` under Terraform's `resource` for undefined `Elem`s:

https://github.com/hashicorp/terraform-schema/pull/174/commits/442c5c891f5798d4a345f8ff55ba1ccb90fec129#diff-f602b672b44eda05c025b0990196bf105681dc20c7db03ff3fd18dd15fcb1b04R103-R107

### DIFF
--- a/schema/constraint_list.go
+++ b/schema/constraint_list.go
@@ -28,16 +28,19 @@ func (List) isConstraintImpl() constraintSigil {
 }
 
 func (l List) FriendlyName() string {
-	elemName := l.Elem.FriendlyName()
-	if elemName != "" {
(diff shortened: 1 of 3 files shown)


dbanck (approved):
LGTM 👍

PR #193, opened 2023-01-23
decoder/schema: Add support for `OneOf` as `Constraint`

radeksimko:

--- 

## TODO

 - [x] **completion**: tests
 - [x] **hover**: tests
 - [x] **semantic tokens**: tests
 - [x] **reference origins**: tests
 - [x] **reference origins**: should we only consider first non-empty constraint, rather than collect all?
 - [x] **reference targets**: tests
 - [x] **reference targets**: should we only consider first non-empty constraint, rather than collect all?

### DIFF
--- a/decoder/expr_one_of.go
+++ b/decoder/expr_one_of.go
@@ -15,11 +15,6 @@ type OneOf struct {
 	pathCtx *PathContext
 }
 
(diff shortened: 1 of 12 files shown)


radeksimko (commented):
LGTM, but I guess you would have to approve the PR 😄 
Screenshot 2023-02-23 at 10 04 21

dbanck (approved):
image

PR #195, opened 2023-01-25
decoder: Add support for `List` as `Constraint`

radeksimko:
Part of https://github.com/hashicorp/terraform-ls/issues/496

Depends on https://github.com/hashicorp/hcl-lang/pull/196

--- 

## TODO

 - [x] **completion**: `schema` empty completion data
 - [x] **completion**: handling of multiline elements (+tests)
 - [x] **completion**: tests
 - [x] **hover**: tests
 - [x] **semantic tokens**: tests
 - [x]  **completion**: review existing tests
 - [x] **hover**: review existing tests
 - [x] **semantic tokens**: review existing tests

---

## Later
 - [ ] **reference targets**: Decide how to collect parent target for the whole list
 - [ ] **reference origins**: tests
 - [ ] **reference targets**: tests

PR #196, opened 2023-01-25
decoder: Adjust `Expression.ReferenceTargets()` address arguments

radeksimko:
As discovered while implementing this method for `List`

 - https://github.com/hashicorp/hcl-lang/pull/195

we will need to pass around an address which may not necessarily refer to the attribute as whole, but different expressions depending on levels of nesting.

This is expected to be useful for most other constraints which can have reference targets inside (e.g. `Object` or `Map`).

This also adds some compilation-time tests to ensure that we don't accidentally change interface somewhere and "turn off" collection of origins or targets.

### DIFF
--- a/decoder/expr_any.go
+++ b/decoder/expr_any.go
@@ -35,7 +35,7 @@ func (a Any) ReferenceOrigins(ctx context.Context, allowSelfRefs bool) reference
 	return nil
 }
 
-func (a Any) ReferenceTargets(ctx context.Context, attrAddr *schema.AttributeAddrSchema) reference.Targets {
+func (a Any) ReferenceTargets(ctx context.Context, addr lang.Address, addrCtx AddressContext) reference.Targets {
 	// TODO
 	return nil
 }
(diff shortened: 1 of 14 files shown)


dbanck (approved):
LGTM, while mildly confusing

We use address as a name for a lot of things: `AttributeAddrSchema`, `lang.Address`, `schema.Address`

radeksimko:
> We use address as a name for a lot of things: AttributeAddrSchema, lang.Address, schema.Address

Yeah, fair point. I think it was an act of balancing "easy to type" vs "easy to read" vs meaning. We could possibly call `lang.Address` => `lang.ResolvedAddress` and `schema.Address` => `schema.AddressSchema`, but I can't think of any simple way to get rid of the "address" in those.

For `AttributeAddrSchema` we could play with some combinations of _target_, or _targetable_, but nothing better comes up to mind right now.

All in all I agree it's not ideal, we can revisit later.

PR #199, opened 2023-01-30
decoder: Add support for `Map` as `Constraint`

radeksimko:

--- 

Existing tests were updated as part of https://github.com/hashicorp/hcl-lang/commit/ddf691aa6ef92495344a38e7c1432e37ae78a964 (not part of this PR) and yield either comparable or better results, except for reference collection, which:

 - we agreed to leave for a separate PR
 - involves other constraints which aren't implemented yet

---


decoder/expr_map_completion_test.go
```diff
@@ -0,0 +1,996 @@
...
+							Keyword: "kw",
+						},
+					},
+				},
+			},
+			`attr = {  foo = kw }
+`,
+			hcl.Pos{Line: 1, Column: 10, Byte: 9},
+			lang.CompleteCandidates([]lang.Candidate{}),
+		},
```
radeksimko, line 350: We could still update the implementation to yield suggestions in this context.

decoder/expr_map_completion_test.go
```diff
@@ -0,0 +1,996 @@
...
+							Keyword: "kw",
+						},
+					},
+				},
+			},
+			`attr = { foo = kw  }
+`,
+			hcl.Pos{Line: 1, Column: 19, Byte: 18},
+			lang.CompleteCandidates([]lang.Candidate{}),
+		},
```
radeksimko, line 398: We could still update the implementation to yield suggestions in this context.

decoder/expr_map_completion_test.go
```diff
@@ -0,0 +1,996 @@
...
+							Keyword: "kw",
+						},
+					},
+				},
+			},
+			`attr = { foo = kw,   bar = kw }
+`,
+			hcl.Pos{Line: 1, Column: 20, Byte: 19},
+			lang.CompleteCandidates([]lang.Candidate{}),
+		},
```
radeksimko, line 446: We could still update the implementation to yield suggestions in this context.

decoder/expr_map_completion_test.go
```diff
@@ -0,0 +1,996 @@
...
+					},
+				},
+			},
+			`attr = {
+  foo = keyword 
+}
+`,
+			hcl.Pos{Line: 2, Column: 17, Byte: 25},
+			lang.CompleteCandidates([]lang.Candidate{}),
+		},
```
radeksimko, line 896: We could still update the implementation to yield suggestions in this context.

decoder/expr_map_completion_test.go
```diff
@@ -0,0 +1,996 @@
...
+
+func TestCompletionAtPos_exprTypeDeclaration(t *testing.T) {
+	testCases := []struct {
+		testName           string
+		attrSchema         map[string]*schema.AttributeSchema
+		cfg                string
+		pos                hcl.Pos
+		expectedCandidates lang.Candidates
+	}{
+		{
```
dbanck, line 23: We could add one more test case with an empty configuration, like

```
`
`,
```
to test the full attribute completion `attr = {\n ${1}\n}`

dbanck (approved):
Well done! I think everything is working as intended.

Just a small suggestion for another test case.

decoder/expr_map_completion_test.go
radeksimko: We can, but that is then testing a different code path, so not the same "unit" as all the other tests.

radeksimko:
@dbanck I have made some last minor changes around the handling of the edge cases, as previously mentioned and also fixed one minor bug I found while working on that.

PTAL

decoder/expr_map_completion_test.go
dbanck: Oh, right. Please ignore my suggestion

dbanck (approved):
LGTM! Thanks for taking the time to update the completion

PR #202, opened 2023-02-10
Fix `LocalAddr` creation in target collection for JSON files

dbanck:
Sometimes, we create local addresses with `nil` steps, and the LS crashes when iterating over those. This PR aims to fix that.

We don't get a body range for JSON-based configuration files, so we can never set the `TargetableFromRangePtr` for collected targets. 

https://github.com/hashicorp/hcl-lang/blob/4804a1caffd2d165978b0dc75ca4df6bd6ada14e/decoder/reference_targets.go#L687-L689

Without this pointer, we can't build a local address. So instead of trying to be clever and set a specific length, we now create all local addresses with a zero length.

Fixes https://github.com/hashicorp/vscode-terraform/issues/1338
Fixes https://github.com/hashicorp/terraform-ls/issues/1157

### DIFF
--- a/decoder/reference_targets.go
+++ b/decoder/reference_targets.go
@@ -740,7 +740,7 @@ func (d *PathDecoder) collectInferredReferenceTargetsForBody(addr lang.Address, 
 
 		blockRef := reference.Target{
 			Addr:        blockAddr,
-			LocalAddr:   make(lang.Address, len(selfRefAddr)),
+			LocalAddr:   make(lang.Address, 0),
 			ScopeId:     bAddrSchema.ScopeId,
 			Type:        cty.Object(bodySchemaAsAttrTypes(bCollection.Schema.Body)),
 			Description: bCollection.Schema.Description,
@@ -748,6 +748,7 @@ func (d *PathDecoder) collectInferredReferenceTargetsForBody(addr lang.Address, 
 			RangePtr:    blk.Range.Ptr(),
 		}
 		if bAddrSchema.DependentBodySelfRef && selfRefBodyRangePtr != nil {
+			blockRef.LocalAddr = make(lang.Address, len(selfRefAddr))
 			copy(blockRef.LocalAddr, selfRefAddr)
 			blockRef.LocalAddr = append(blockRef.LocalAddr, lang.AttrStep{Name: bType})
 			blockRef.TargetableFromRangePtr = selfRefBodyRangePtr.Ptr()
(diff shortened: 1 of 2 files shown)


decoder/reference_targets.go
```diff
@@ -740,14 +740,15 @@ func (d *PathDecoder) collectInferredReferenceTargetsForBody(addr lang.Address,
...
-			LocalAddr:   make(lang.Address, len(selfRefAddr)),
+			LocalAddr:   make(lang.Address, 0),
 			ScopeId:     bAddrSchema.ScopeId,
 			Type:        cty.Object(bodySchemaAsAttrTypes(bCollection.Schema.Body)),
 			Description: bCollection.Schema.Description,
 			DefRangePtr: blk.DefRange.Ptr(),
 			RangePtr:    blk.Range.Ptr(),
 		}
 		if bAddrSchema.DependentBodySelfRef && selfRefBodyRangePtr != nil {
+			blockRef.LocalAddr = make(lang.Address, len(selfRefAddr))
```
dbanck, line 751: I wonder if it would make more sense to keep creating the slice with a specific length and clear it instead, like:

```go
blockRef := reference.Target{
	Addr:        blockAddr,
	LocalAddr:   make(lang.Address, len(selfRefAddr)),
	...
}
if bAddrSchema.DependentBodySelfRef && selfRefBodyRangePtr != nil {
	copy(blockRef.LocalAddr, selfRefAddr)
	blockRef.LocalAddr = append(blockRef.LocalAddr, lang.AttrStep{Name: bType})
	blockRef.TargetableFromRangePtr = selfRefBodyRangePtr.Ptr()
} else {
	blockRef.LocalAddr = lang.Address{}
}
```

This would keep the "hot" path as is and only modify the slice, when dealing with JSON files.

radeksimko: I'm not sure what's better.

The main motivation with precise length is reduce memory consumption, but I'm guessing that is thwarted anyway with the use of `append()` which will likely grow the allocation by more than what's needed for 1 element anyway? Not sure about that.

Also I am not sure what happens when a pre-allocated slice is set to an empty one, I'm assuming GC will take care of it, but the memory would have been allocated before subsequently being freed.

So unless we're looking to do performance optimisations - which I don't think was the goal of this PR - then I don't think it matters.

radeksimko (approved):
The crash was related to a different PR which I had accidentally checked out and forgot about https://github.com/hashicorp/terraform-schema/pull/174/files

it causes the crash on `main` as well, so it doesn't seem affected by this PR.

--- 

One more thing I forgot to mention in my previous review is that _if_ it makes things easier, we could gate `LocalAddr` setting on `hclsyntax` / JSON, since we won't ever need local addresses inside JSON anyway, since we don't do completion _inside_ JSON files anyway.

decoder/reference_targets.go
dbanck: 👍 I'll keep the changes as they are as I'm not doing any performance optimizations.

dbanck:
I think gating `LocalAddr` wouldn't make the code considerably easier to read, so I would keep it as is. Happy to revisit this in the future.

PR #206, opened 2023-02-14
Add copywrite config

jpogran:


.copywrite.hcl
```diff
@@ -0,0 +1,10 @@
+schema_version = 1
+
+project {
+  license        = "MPL-2.0"
```
radeksimko, line 4:
```suggestion
  license = "MPL-2.0"
```
If only https://github.com/hashicorp/vscode-hcl/issues/98 was a thing 😄

radeksimko (approved):

PR #207, opened 2023-02-15
Add function signature types

dbanck:
This PR extracts the function signature types from #135 to enable us to implement upstream changes in terraform-schema before completing the function signatures features as a whole.

We can use the new signature types to embed function signatures in terraform-schema.

---

* Part of https://github.com/hashicorp/terraform-ls/issues/37

### DIFF
--- a/decoder/path_context.go
+++ b/decoder/path_context.go
@@ -16,8 +16,7 @@ type PathContext struct {
 	ReferenceOrigins reference.Origins
 	ReferenceTargets reference.Targets
 	Files            map[string]*hcl.File
-
-	// TODO: Functions
+	Functions        map[string]schema.FunctionSignature
 }
 
 type pathCtxKey struct{}
--- a/schema/signature.go
+++ b/schema/signature.go
@@ -1,0 +1,24 @@
+package schema
+
+import (
+	"github.com/zclconf/go-cty/cty"
+	"github.com/zclconf/go-cty/cty/function"
+)
+
+type FunctionSignature struct {
+	// Description is an optional human-readable description
+	// of the function.
+	Description string
+
+	// ReturnType is the ctyjson representation of the function's
+	// return types based on supplying all parameters using
+	// dynamic types. Functions can have dynamic return types.
+	ReturnType cty.Type
+
+	// Params describes the function's fixed positional parameters.
+	Params []function.Parameter
+
+	// VarParam describes the function's variadic
+	// parameter if it is supported.
+	VarParam *function.Parameter
+}


schema/signature.go
```diff
@@ -0,0 +1,24 @@
...
+	// return types based on supplying all parameters using
+	// dynamic types. Functions can have dynamic return types.
+	ReturnType cty.Type
+
+	// Parameters describes the function's fixed positional parameters.
+	Params []function.Parameter
+
+	// VariadicParameter describes the function's variadic
+	// parameter if it is supported.
+	VarParam *function.Parameter
```
radeksimko, line 23:
```suggestion
	// Params describes the function's fixed positional parameters.
	Params []function.Parameter

	// VarParam describes the function's variadic
	// parameter if it is supported.
	VarParam *function.Parameter
```

I'm not sure if the short names are better here, but I guess they align with the relevant function names in go-cty, so at least we're consistent 🤷🏻 

 - https://pkg.go.dev/github.com/zclconf/go-cty@v1.12.1/cty/function#Function.Params
 - https://pkg.go.dev/github.com/zclconf/go-cty@v1.12.1/cty/function#Function.VarParam

schema/signature.go
```diff
@@ -0,0 +1,24 @@
+package schema
+
+import (
+	"github.com/zclconf/go-cty/cty"
+	"github.com/zclconf/go-cty/cty/function"
+)
+
+type FuncSignature struct {
```
radeksimko, line 8:
```suggestion
type FunctionSignature struct {
```

This would IMO better align with `function.Function` in `go-cty`: https://pkg.go.dev/github.com/zclconf/go-cty@v1.12.1/cty/function#Function

decoder/path_context.go
```diff
@@ -16,8 +16,7 @@ type PathContext struct {
 	ReferenceOrigins reference.Origins
 	ReferenceTargets reference.Targets
 	Files            map[string]*hcl.File
-
-	// TODO: Functions
+	Functions        map[string]schema.FuncSignature
```
radeksimko, line 19:
```suggestion
	Functions        map[string]schema.FunctionSignature
```

lang/signature.go
```diff
@@ -0,0 +1,24 @@
...
+	ActiveParameter uint
+}
+
+type FuncParameter struct {
+	Name string
+
+	// Description is an optional human-readable description
+	// of the parameter.
+	Description MarkupContent
+}
```
radeksimko, line 24: None of the `lang.*` types seem currently required for https://github.com/hashicorp/terraform-schema/pull/145 AFAICT, so shall we leave it for a separate PR which introduces the code that needs it? e.g. signatureHelp method?

dbanck: Yes, we can do that.

radeksimko (approved):

PR #208, opened 2023-02-17
decoder: Adjust `Constraint`.`ReferenceTargets` method args

radeksimko:
The new interface better reflects a scenario where expression can be targeted by itself, but not the attribute. In Terraform this may occur in `required_providers` > `configuration_aliases` where entries are targetable provider names, but configuration_aliases attribute is not.

In that ^ scenario the expected schema would be e.g.

```go
"configuration_aliases": &schema.AttributeSchema{
	Constraint: schema.Set{
		Elem: schema.Reference{
				Address: &schema.ReferenceAddrSchema{
					ScopeId: refscope.ProviderScope,
				},
				Name: "provider",
			},
		},
	},
},
```
i.e. there would be no `Address` set on the `AttributeSchema`.

I renamed `AddressContext` to `TargetContext` as per earlier discussion in a different PR about lots of things called "address". Hopefully this is a better name 🤞🏻 

Finally, the change in moving resolved attribute address into the `TargetContext` reflects the existing assumption that there's no point in collecting targets within the expression if the address is unresolvable and reduces the number of arguments down to 2.

### DIFF
--- a/decoder/expression.go
+++ b/decoder/expression.go
@@ -50,8 +50,8 @@ type TargetContext struct {
 	// is addressable as a type-less reference
 	AsReference bool
 
-	// AttributeAddress represents resolved address for the attribute
-	// which the expression belongs to.
+	// AttributeAddress represents a resolved address for the attribute
+	// to which the expression belongs.
 	AttributeAddress lang.Address
 }
 
--- a/decoder/expression.go
+++ b/decoder/expression.go
@@ -24,11 +24,11 @@ type ReferenceOriginsExpression interface {
 }
 
 type ReferenceTargetsExpression interface {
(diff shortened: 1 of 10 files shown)


decoder/expression.go
```diff
@@ -49,6 +49,10 @@ type AddressContext struct {
 	// AsReference defines whether the attribute
 	// is addressable as a type-less reference
 	AsReference bool
+
+	// AttributeAddress represents resolved address for the attribute
+	// which the expression belongs to.
```
dbanck, line 54:
```suggestion
	// AttributeAddress represents a resolved address for the attribute
	// to which the expression belongs.
```

dbanck (approved):
LGTM, just a very minor nit

PR #210, opened 2023-02-17
schema: Introduce `EmptyHoverData()` to `Constraint`

radeksimko:
Similar to the existing `EmptyCompletionData()`, in some contexts it feels more appropriate to display something more helpful than just short "friendly name" for the type.

The friendly name is still relevant to display e.g. next to the completion item, where there is no space for more details.

We could potentially use this also when displaying hover for attributes, where we currently use just `FriendlyName()`, but I'd leave that for another discussion later.

The aim is to be able to mimic the existing hover data for complex `LiteralType` constraints.

I also used the pointer (`*HoverData`) just so that we could also still decide at runtime in each constraint whether or not to return hover data. This is because e.g.`List` or `Set` may behave differently depending on the Constraints of its elements.

Similar logic could in theory apply to `EmptyCompletionData()` - i.e. one could argue we should also be returning a pointer there, but the added complexity there is that we also use that when pre-filling required fields and in that context we don't really have any reasonable fallback, such as `FriendlyName()`. We could fall back to just empty string, but I'm not sure that's always appropriate as we would end up generating grossly incomplete configuration with lots of trailing `=` which may be harder to navigate and produce a lot of diagnostics. Maybe the ideal solution here is to have another method for pre-filling required fields. 🤷🏻

### DIFF
--- a/schema/constraint.go
+++ b/schema/constraint.go
@@ -1,6 +1,9 @@
 package schema
 
-import "github.com/zclconf/go-cty/cty"
+import (
+	"github.com/hashicorp/hcl-lang/lang"
+	"github.com/zclconf/go-cty/cty"
+)
 
 type constraintSigil struct{}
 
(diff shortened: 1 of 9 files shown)


dbanck (approved):
LGTM 👍 

I appreciate the additional context in the description!

PR #211, opened 2023-02-17
decoder: Introduce helper functions for completion inside `ObjectConsExpr`

radeksimko:
This PR introduces two small helper functions decoupled from a few discovered use cases / PRs:

 - https://github.com/hashicorp/hcl-lang/pull/183 (for `object({ ... })`)
 - https://github.com/hashicorp/hcl-lang/pull/186 (through `Object` or `Map`)
 - https://github.com/hashicorp/hcl-lang/pull/199
 - https://github.com/hashicorp/hcl-lang/pull/203
 - it's also expected to be useful in `AnyExpression` constraint

The last attached test case illustrates an example of the problem.

--- 

### Implementation Notes

Our use cases don't _actually_ involve multi-byte runes, so in theory we could be just comparing single bytes and avoid the decoding, but that then makes the consumer side a bit more verbose since `'\n'` would need to be written e.g. as `byte('\n')`. We could technically allow runes and treat them like single bytes, but that just feels wrong.

Since we'd usually seek through relatively small slice, the decoding is unlikely to pose performance issue, but we can revisit the solution if it does.

### DIFF
--- a/decoder/expression.go
+++ b/decoder/expression.go
@@ -197,3 +197,10 @@ func recoverLeftBytes(b []byte, pos hcl.Pos, f func(byteOffset int, r rune) bool
 
 	return []byte{}
 }
+
+// isObjectItemTerminatingRune returns true if the given rune
+// is considered a left terminating character for an item
+// in hclsyntax.ObjectConsExpr.
+func isObjectItemTerminatingRune(r rune) bool {
+	return r == '\n' || r == ',' || r == '{'
+}
--- a/decoder/expression.go
+++ b/decoder/expression.go
@@ -2,6 +2,7 @@ package decoder
 
 import (
 	"context"
+	"unicode/utf8"
 
 	"github.com/hashicorp/hcl-lang/lang"
 	"github.com/hashicorp/hcl-lang/reference"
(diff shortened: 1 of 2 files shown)


decoder/expression.go
```diff
@@ -167,3 +168,39 @@ func newEmptyExpressionAtPos(filename string, pos hcl.Pos) hcl.Expression {
...
+// This allows recovery of incomplete configuration which is not
+// present in the parsed AST during completion.
+//
+// Zero bytes is returned if no match was found.
+func recoverLeftBytes(b []byte, pos hcl.Pos, f func(byteOffset int, r rune) bool) []byte {
+	lastRune, size := utf8.DecodeLastRune(b[:pos.Byte])
+	offset := pos.Byte - size
+
+	// check for early match
+	if f(pos.Byte, lastRune) {
```
dbanck, line 183:
```suggestion
	firstRune, size := utf8.DecodeLastRune(b[:pos.Byte])
	offset := pos.Byte - size

	// check for early match
	if f(pos.Byte, firstRune) {
```

Nit: Isn't this the first rune in the string we're getting?

dbanck (approved):
LGTM! Just a small suggestion, feel free to merge anyway.

decoder/expression.go
radeksimko: It depends on the perspective I suppose 😄 It's the first one when counting bytes from right to left, it's the last one when decoding the run (from left to right).

PR #213, opened 2023-02-21
decoder: Introduce rawObjectKey to extract raw keys

radeksimko:
As per in-line comment:

```go
// rawObjectKey extracts raw key (as string) from KeyExpr of
// any hclsyntax.ObjectConsExpr along with the corresponding range
// and boolean indicating whether the extraction was successful.
```

This is expected to be used in any expression type that involves `ObjectConsExpr`, i.e.

 - https://github.com/hashicorp/hcl-lang/pull/203
 - https://github.com/hashicorp/hcl-lang/pull/199
 - https://github.com/hashicorp/hcl-lang/pull/183

for validating keys before reporting them as semantic tokens, or to isolate the unquoted part which in turn enables validation/lookup of the attribute name in case of an object. Range then allows us to also extract the prefix for any necessary filtering when doing completion for the whole attribute.

### DIFF
--- a/decoder/expression.go
+++ b/decoder/expression.go
@@ -204,3 +204,49 @@ func recoverLeftBytes(b []byte, pos hcl.Pos, f func(byteOffset int, r rune) bool
 func isObjectItemTerminatingRune(r rune) bool {
 	return r == '\n' || r == ',' || r == '{'
 }
+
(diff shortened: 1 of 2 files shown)


dbanck (approved):
👍 I like the tests for that help

PR #214, opened 2023-02-21
decoder/TargetContext: AttributeAddress -> ParentAddress

radeksimko:
Whilst trying to use the new interface, I ran into an awkward situation which I anticipate to be common between all the complex types (`Object`, `Map`, `Set`, `List`, `Tuple`).

For simple types (bool, string, number, ...) which are targetable by themselves, the meaning of `ParentAddress` will indeed be "attribute address", but for the complex ones, the expectation is for the relevant expression type to construct the "intermediate" parent address for each element/item and _then_ hand it over to the underlying element/item expression with that context.

Also, in order to append item/element parts of the address safely (without modifying the original context), a `Copy()` function is needed.


dbanck (approved):
LGTM. I'm just unsure if there might be a better name than `ParentAddress`, since it may only be the address of the attribute in some cases, but I can't think of one right now.

PR #217, opened 2023-02-22
schema: Implement `ConstraintType()` for all remaining constraints

radeksimko:
We could add tests, but in all cases it's either simple pass of an existing type (1LOC) or just putting the same data into a different shape, so I'm not sure it's worth testing in isolation.

We can still test it indirectly as part of reference target collection.


radeksimko:
Ooops, forgot about `OneOf`

dbanck (approved):
LGTM 👍 

Should we add a hint to `OneOf`, that we pick the first type aware constraint?

PR #218, opened 2023-02-22
schema: Introduce `nestingLevel` to EmptyCompletionData()

radeksimko:
While trying to implement `EmptyCompletionData()` for `Object` and `Map`, I realised that in both cases the constraint may be nested, i.e. the object may be nested under another object/map. In that scenario we need to render the right indentation depending on the nesting level we're at.

The updated interface accounts for this.


dbanck (approved):
👍

PR #220, opened 2023-02-23
schema: Implement `EmptyHoverData()` for all constraints

radeksimko:
The implementation for `LiteralValue` specifically turns out to be a little more involved than I originally hoped, but I decided to include it in the PR as well.

It reminds me that implementing https://github.com/hashicorp/hcl-lang/pull/187 may be more involved in general, but we can cross the bridge when we come to it.

The primary motivation to get this PR out was to enable more comprehensive testing of hover in https://github.com/hashicorp/hcl-lang/pull/203

### DIFF
--- a/schema/constraint_literal_value.go
+++ b/schema/constraint_literal_value.go
@@ -1,6 +1,11 @@
 package schema
 
 import (
+	"fmt"
+	"sort"
+	"strconv"
+	"strings"
+
 	"github.com/hashicorp/hcl-lang/lang"
 	"github.com/zclconf/go-cty/cty"
 )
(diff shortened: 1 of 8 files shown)


schema/constraint_object.go
```diff
@@ -45,8 +49,72 @@ func (o Object) EmptyCompletionData(placeholder int, nestingLevel int) Completio
...
+		return []string{}
+	}
+
+	constraints := attributes
+	names := make([]string, len(constraints))
+	i := 0
+	for name := range constraints {
+		names[i] = name
+		i++
+	}
```
dbanck, line 114:
```suggestion
	names := make([]string, 0, len(constraints))
	for name := range constraints {
		names = append(names, name)
	}
```

Nit: I think using `append` here makes it more readable.

schema/constraint_literal_value.go
```diff
@@ -39,10 +44,171 @@ func (lv LiteralValue) EmptyCompletionData(nextPlaceholder int, nestingLevel int
...
+		return []string{}
+	}
+
+	constraints := valueMap
+	names := make([]string, len(constraints))
+	i := 0
+	for name := range constraints {
+		names[i] = name
+		i++
+	}
```
dbanck, line 194:
```suggestion
	names := make([]string, 0, len(constraints))
	for name := range constraints {
		names = append(names, name)
	}
```

Nit: I think using `append` here makes it more readable.

dbanck (approved):
LGTM; just a minor stylistic suggestion. Feel free to merge anyway.

I'm unsure if hovering on primitive types provides much value to a user since the code is already directly visible in the editor, and the hover dialog repeats it. But let's have this conversation another day.

schema/constraint_object.go
radeksimko: Even though the performance (memory) benefit is _probably_ minor, I'd say the readability gain is negligible too, so I'd prefer to keep this as is and benefit from lower memory consumption.

radeksimko:
> I'm unsure if hovering on primitive types provides much value to a user since the code is already directly visible in the editor, and the hover dialog repeats it. But let's have this conversation another day.

I don't disagree, but it also feels like unnecessary complexity atm to want to explicitly block that hover data in that particular context, when we still need it when the primitive type is enclosed in e.g. object. Happy to discuss that later though!

schema/constraint_object.go
dbanck: Sure, we can keep this as is.

---

Yet, the memory consumptions for both approaches should be very similar. `make([]string, 0, len(constraints))` will initialize the slice with a zero-length and a capacity of `len(constraints)`, so Go internally already preallocates all the required elements. Therefore, adding up to `len(constraints)` elements to the slice means reusing the same backing array and no further allocations. The only difference is the repeated calls to the built-in `append` function, which has a small overhead compared to a direct assignment. Or am I missing something?

PR #222, opened 2023-02-24
schema: Implement `EmptyCompletionData()` for all constraints

radeksimko:
## Testing

Aside from reviewing the attached unit tests, this can also be tested when combined with https://github.com/hashicorp/terraform-schema/pull/174

There are two contexts in which `EmptyCompletionData()` is used, as shown below.

### DIFF
--- a/schema/constraint_tuple.go
+++ b/schema/constraint_tuple.go
@@ -46,7 +46,7 @@ func (t Tuple) EmptyCompletionData(nextPlaceholder int, nestingLevel int) Comple
 		return CompletionData{
 			NewText:         "[]",
 			Snippet:         fmt.Sprintf("[ ${%d} ]", nextPlaceholder),
-			LastPlaceholder: nextPlaceholder + 1,
+			NextPlaceholder: nextPlaceholder + 1,
 		}
 	}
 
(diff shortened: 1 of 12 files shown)


schema/constraint_tuple.go
```diff
@@ -42,8 +42,38 @@ func (t Tuple) Copy() Constraint {
...
+		if cData.NewText == "" || cData.Snippet == "" {
+			return CompletionData{
+				NewText:         "[]",
+				Snippet:         fmt.Sprintf("[ ${%d} ]", nextPlaceholder),
+				TriggerSuggest:  cData.TriggerSuggest,
+				NextPlaceholder: nextPlaceholder + 1,
+			}
+		}
+		elemNewText[i] = cData.NewText
+		elemSnippets[i] = cData.NewText
```
dbanck, line 68:
```suggestion
		elemSnippets[i] = cData.Snippet
```

dbanck (approved):
LGTM 👍 But one small typo

schema/constraint_tuple.go
radeksimko: Good catch!

PR #231, opened 2023-03-15
decoder: fix Map test name

radeksimko:
This fixes an earlier copy-paste error similar to https://github.com/hashicorp/hcl-lang/pull/223


dbanck (approved):
_takes mental note to better review test names in the future_

PR #232, opened 2023-03-15
decoder: Add support for `AnyExpression` as `Constraint`

dbanck:
* Depends on https://github.com/hashicorp/hcl-lang/pull/246

--- 

## UX
We provide completion of references, function and literals for all expressions

Completion, hover and semantic tokens also work for nested expressions, like within objects or function calls.

---

* Part of https://github.com/hashicorp/terraform-ls/issues/496
* Part of https://github.com/hashicorp/terraform-ls/issues/37


radeksimko (approved):
:shipit:

Issue #228, opened 2023-03-13
decoder: Finish reference collection implementation for `Constraint`


radeksimko:
## Context

For the following constraints:

 - https://github.com/hashicorp/hcl-lang/pull/194
 - https://github.com/hashicorp/hcl-lang/pull/195
 - https://github.com/hashicorp/hcl-lang/pull/198
 - https://github.com/hashicorp/hcl-lang/pull/199
 - https://github.com/hashicorp/hcl-lang/pull/203

we have implemented reference collection to our best effort/knowledge, but were unable to test most of that logic due to inability to use other "collectable" constraints, such as

 - https://github.com/hashicorp/hcl-lang/pull/187
 - https://github.com/hashicorp/hcl-lang/pull/186
 - https://github.com/hashicorp/hcl-lang/pull/185

This is therefore a reminder to come back to all constraints and ensure that the implementation is indeed correct and tested properly.

## Proposal

 - [ ] **reference origins**: `List` - add tests
 - [ ] **reference targets**: `List` - collect parent target for the whole list
 - [ ] **reference targets**: `List` - add tests
 - [ ] **reference origins**: `Set` - add tests
 - [ ] **reference targets**: `Set` - collect parent target for the whole set
 - [ ] **reference targets**: `Set` - add tests
 - [ ] **reference origins**: `Tuple` - add tests
 - [ ] **reference targets**: `Tuple` - collect parent target for the whole tuple
 - [ ] **reference targets**: `Tuple` - add tests
 - [ ] **reference origins**: `Map` - add tests
 - [ ] **reference targets**: `Map` - add tests
 - [ ] **reference origins**: `Object` - add tests
 - [ ] **reference targets**: `Object` - add tests
 - [ ] **reference targets**: `LiteralType` - add tests
 - [ ] **reference targets**: `LiteralValue` - add tests

PR #237, opened 2023-03-20
Finish implementation of reference target collection for List/Set/Tuple

radeksimko:
This partially addresses https://github.com/hashicorp/hcl-lang/issues/228


radeksimko:
Replaced by https://github.com/hashicorp/hcl-lang/pull/239

PR #239, opened 2023-03-20
Finish reference target collection for all constraints

radeksimko:
Depends on https://github.com/hashicorp/hcl-lang/pull/241

Closes https://github.com/hashicorp/hcl-lang/issues/228

This excludes `LiteralValue`, but [we don't actually use that constraint in reference target context currently in `terraform-schema` anyway](https://github.com/search?q=repo%3Ahashicorp%2Fterraform-schema%20LegacyLiteralValue&type=code).

--- 

## UX

(requires https://github.com/hashicorp/hcl-lang/pull/241 and https://github.com/hashicorp/terraform-schema/pull/174 )


jpogran (approved):
:shipit:

PR #244, opened 2023-04-05
Replace extension schema `Expr` w/ `Constraint`

dbanck:
Depends on https://github.com/hashicorp/hcl-lang/pull/232


radeksimko (approved):
LGTM 🚀

PR #246, opened 2023-04-05
schema: Implement `EmptyCompletionData` for `AnyExpression`

radeksimko:
I'm not sure if this is the best solution, but it mostly mimics the existing behaviours of `TraversalExpr` (legacy `Reference`) and `LiteralTypeExpr` (legacy `LiteralType`).

That is, we produce pre-filling, when enabled:

the (perhaps undesirable?) side effect though is that we also pre-fill on attribute completion:

where it might be more appropriate to trigger completion and bring up functions/references:

--- 

So one better solution I can think of would involve attaching another context value somewhere here https://github.com/hashicorp/hcl-lang/blob/fba9786f3c123b25fd1d213716e385cb9b5b36a9/decoder/attribute_candidates.go#L21-L24 and add that to the condition, such that for `LiteralType`-like behaviour (pre-filling empty values) we not only expect the prefilling to be toggled on, but also to _not_ be providing completion for an attribute.

### DIFF
--- a/schema/constraint_any_expression.go
+++ b/schema/constraint_any_expression.go
@@ -34,8 +34,20 @@ func (ae AnyExpression) Copy() Constraint {
 }
 
 func (ae AnyExpression) EmptyCompletionData(ctx context.Context, nextPlaceholder int, nestingLevel int) CompletionData {
-	// TODO
-	return CompletionData{}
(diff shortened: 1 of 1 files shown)


dbanck (approved):
👍 I think mimicking the existing behavior is fine for now

PR #247, opened 2023-04-05
Collect implied reference targets for complex types

radeksimko:
## Context

Previously, as part of https://github.com/hashicorp/hcl-lang/pull/239 we forgot about the fact that complex expression types, such as `List`, `Set`, `Tuple`, `Map` or `Object` may end up being collected as reference targets in a context where there is no corresponding config/expression.

This is currently surfaced in the UX by references to any complex types missing unless explicitly declared:

## Proposal (fix)

- [x] Collect `List` targets for empty expression
  - [x] Cover with tests (AsExprType + AsReference)
- [x] Collect `Set` targets for empty expression
  - [x] Cover with tests (AsExprType + AsReference)
- [x] Collect `Map` targets for empty expression
  - [x] Cover with tests (AsExprType + AsReference)
- [x] Collect `Tuple` targets for empty expression (including nested expressions for each element in `Elems`)
  - [x] Cover with tests (AsExprType + AsReference)
- [x] Collect `Object` targets for empty expression (including nested expressions for each attribute in `Attributes`)
  - [x] Cover with tests (AsExprType + AsReference)

### DIFF
--- a/decoder/expr_list_ref_targets.go
+++ b/decoder/expr_list_ref_targets.go
@@ -14,6 +14,10 @@ import (
 )
 
 func (list List) ReferenceTargets(ctx context.Context, targetCtx *TargetContext) reference.Targets {
+	if isEmptyExpression(list.expr) && targetCtx != nil {
+		return list.wholeListReferenceTargets(targetCtx, nil)
+	}
+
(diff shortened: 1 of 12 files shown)


dbanck (approved):
:shipit:

PR #251, opened 2023-04-12
Update legacy tests to use `Constraint`

radeksimko:
## TODO

 - [x] https://github.com/hashicorp/hcl-lang/pull/253
 - [x] https://github.com/hashicorp/hcl-lang/pull/255
 - [x] https://github.com/hashicorp/hcl-lang/pull/254
 - [x] https://github.com/hashicorp/hcl-lang/pull/259
 - [x] https://github.com/hashicorp/hcl-lang/pull/257

PR #253, opened 2023-04-13
decoder: Reflect `LiteralValue` `Description` & `IsDeprecated` in completion/hover

radeksimko:
As discovered in https://github.com/hashicorp/hcl-lang/pull/251

`LiteralValue` constraint (prior to this patch) ignores `Description` and `IsDeprecated` fields during completion and hover.


dbanck (approved):
👍

PR #255, opened 2023-04-13
fix(schema): crash when completing `LiteralType{Type: cty.Tuple}`

radeksimko:
As discovered in manual testing and in #251 

Completion of an attribute of `cty.Tuple` type would previously cause a crash.

I also decided to add some more tests to cover the (clearly non-trivial enough to hide a bug) `EmptyCompletionData()` method on `LiteralType`.


dbanck (approved):
Good catch!

PR #256, opened 2023-04-13
decoder: Implement `InferType` for complex expressions

radeksimko:
As discovered in #251 

Related to https://github.com/hashicorp/hcl-lang/pull/247

One alternative (possibly simpler) solution here would be falling back to `ConstraintType()` of the corresponding constraint (List, Set, Tuple, Map, Object). I do see some benefits in the ability to _infer_ types though in the long run.

## UX Impact

There currently aren't any constraints in the Terraform schema involving "partially known" types as explained above, so there's no expected UX impact on Terraform LS users.

This was only discovered because some existing tests in #251 use the mentioned constraints.

### DIFF
--- a/decoder/expr_list.go
+++ b/decoder/expr_list.go
@@ -6,10 +6,59 @@ package decoder
 import (
 	"github.com/hashicorp/hcl-lang/schema"
 	"github.com/hashicorp/hcl/v2"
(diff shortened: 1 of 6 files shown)


radeksimko:
Closing in favour of https://github.com/hashicorp/hcl-lang/pull/259

PR #257, opened 2023-04-13
decoder: Collect targets w/ interpolation for `Any` correctly

radeksimko:
As discovered in #251

In a hypothetical perfect/ideal scenario we could collect just targets which match the type but in reality we cannot always establish the expression type, because:

 - we don't recognise all expression types (yet)
 - we'd have to re-implement DAG in some way to ensure we collect and interpolate in the right order
 - not all variables may be known at editing time (some may be provided only at runtime)

Also it's questionable whether it's beneficial to suppress mismatching targets.
We _mostly_ only do it as a hack to aid `LiteralType` and `Reference` in `OneOf`.

The LOC in the diff may seem high, but 95% of that is a result of copying over all tests for reference target collection from `LiteralType` & replacing `LiteralType{Type}` with `AnyExpression{OfType}` + updating that one test for "mismatching" constraint which becomes a positive test.

## UX Before

Screenshot 2023-04-13 at 15 33 27

## UX After

Screenshot 2023-04-13 at 20 58 09

### DIFF
--- a/decoder/expr_any_ref_targets.go
+++ b/decoder/expr_any_ref_targets.go
@@ -8,17 +8,118 @@ import (
 
 	"github.com/hashicorp/hcl-lang/reference"
 	"github.com/hashicorp/hcl-lang/schema"
+	"github.com/hashicorp/hcl/v2"
+	"github.com/zclconf/go-cty/cty"
 )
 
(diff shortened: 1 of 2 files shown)


dbanck (approved):
👍

PR #259, opened 2023-04-14
decoder: Fix collection of implied declared targets of complex types

radeksimko:
As discovered in https://github.com/hashicorp/hcl-lang/pull/251

## UX Impact

There currently aren't any constraints in the Terraform schema involving _targetable_ `List`/`Set`/`Tuple`/`Map`/`Object` - most occurrences are either not targetable or are implied through `LiteralType` or `AnyExpression`, or part of `OneOf` where `AnyExpression` takes precedence when collecting targets.

TL;DR there's no expected UX impact on Terraform LS users.

This was only discovered because some existing tests being updated in #251

## Implementation Notes

There are situations where the schema may be vague and the user may benefit from seeing the fully inferred type, rather than just `any` in certain contexts. We already do that for the most obvious case of `LiteralType{Type: cty.DynamicPseudoType}` and `AnyExpression{}` equivalent.

https://github.com/hashicorp/hcl-lang/blob/d2c7ba3e0033b84af71dcb98d86f73fa3c110e7b/decoder/expr_literal_type_ref_targets.go#L18-L23

We could potentially implement `InferType()` for all the impacted types, but there's no immediate benefit given the above + this would require some more changes to take advantage of such type inference. We'd e.g. have to introduce `TypeHint` to `TargetContext` and get all the expression types to actually use the inferred type when collecting the targets.

The edge cases this could handle involve either explicit `List`, `Set`, `Tuple`, `Map` or `Object` constraints with `cty.DynamicPseudoType` element type, or `LiteralType{Type: cty.List(cty.DynamicPseudoType)}` and similar.

### DIFF
--- a/decoder/expr_list_ref_targets_test.go
+++ b/decoder/expr_list_ref_targets_test.go
@@ -459,6 +459,73 @@ func TestCollectRefTargets_exprList_implied_hcl(t *testing.T) {
 			},
 		},
 		{
+			"declared as type",
+			&schema.BodySchema{
+				Blocks: map[string]*schema.BlockSchema{
+					"blk": {
+						Address: &schema.BlockAddrSchema{
+							Steps: schema.Address{
+								schema.StaticStep{Name: "blk"},
+							},
+							BodyAsData: true,
+							InferBody:  true,
+						},
+						Body: &schema.BodySchema{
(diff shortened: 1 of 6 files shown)


dbanck (approved):
I like the approach and we can always refine it in the future if necessary.

PR #294, opened 2023-08-02
Schema based validation

jpogran:
- [x] extraneous block labels - e.g. variable "one" "two" {
- [x] missing block labels - e.g. variable {}
- [x] unknown blocks (top level and within resources)
- [x] missing required attributes
- [x] unknown attributes
- [x] missing required blocks (with set minimum instances of block)
- [x] too many blocks (with set maximum instances of block)
- [x] deprecated attributes/blocks (warning)


radeksimko (approved):
I pushed a minor change in the last commit to aid readability but otherwise it LGTM at this point. There's more work to do on readability/composition but we can do that as part of another PR.

PR #299, opened 2023-08-09
Support schema extensions in early validation

jpogran:
This does not raise diagnostics if schema extensions are enabled and Count or ForEach is enabled.

In the work on functions & expressions we introduced the concept of extensions for handling count and for_each. Both aren't part of the regular attributes for a block, but are introduced by enabling the specific extensions for a block.

This adds count and for_each at runtime like completion, hover, and other parts do. Since these are optional attributes, there is are no diagnostics necessary if they are not present.

### DIFF
--- a/decoder/body_candidates.go
+++ b/decoder/body_candidates.go
@@ -8,6 +8,7 @@ import (
 	"sort"
 	"strings"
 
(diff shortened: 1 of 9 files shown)


radeksimko (approved):
LGTM, I have just added two tests for the `dynamic` block, which also confirmed that the third extension works "out of the box" through the existing schema merging logic.

PR #302, opened 2023-08-10
Implement reference validation

jpogran:
Add the ability to use the collected origin and target references in early validation.

Validation funcs will be provided by terraform-ls for now.

### DIFF
--- a/lang/validate.go
+++ b/lang/validate.go
@@ -20,9 +20,7 @@ func (dm DiagnosticsMap) Extend(diagMap DiagnosticsMap) DiagnosticsMap {
 			dm[fileName] = make(hcl.Diagnostics, 0)
 		}
 
-		// TODO: this seems to wipe instead of extend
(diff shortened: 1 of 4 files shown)


lang/validate.go
```diff
@@ -9,4 +9,21 @@ import (
...
+func (dm DiagnosticsMap) Extend(diagMap DiagnosticsMap) DiagnosticsMap {
+	for fileName, diags := range diagMap {
+		_, ok := dm[fileName]
+		if !ok {
+			dm[fileName] = make(hcl.Diagnostics, 0)
+		}
+
+		// TODO: this seems to wipe instead of extend
+		// dm[fileName].Extend(diags)
+		dm[fileName] = append(dm[fileName], diags...)
```
jpogran, line 25: @radeksimko can you take a look at whether this should stay as you intended or should this change stay?

radeksimko (approved):

PR #303, opened 2023-08-21
Implement `Count()` for `DiagnosticsMap`

dbanck:
We need this to check if the length of the diagnostics has changed.

### DIFF
--- a/lang/validate.go
+++ b/lang/validate.go
@@ -25,3 +25,11 @@ func (dm DiagnosticsMap) Extend(diagMap DiagnosticsMap) DiagnosticsMap {
 
 	return dm
 }
+
+func (dm DiagnosticsMap) Count() int {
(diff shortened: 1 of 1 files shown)


lang/validate.go
```diff
@@ -25,3 +25,11 @@ func (dm DiagnosticsMap) Extend(diagMap DiagnosticsMap) DiagnosticsMap {
 
 	return dm
 }
+
+func (dm DiagnosticsMap) Count() int {
```
radeksimko, line 29:
```suggestion
// Count returns the number of diagnostics for all files
func (dm DiagnosticsMap) Count() int {
```

This is just to avoid any doubts about whether this is possibly an equivalent of `len(DiagnosticsMap)`, i.e. number of entries in the map, i.e. files.

PR #304, opened 2023-08-23
Fix `DiagnosticsMap` `Extend`

dbanck:
Before the fix, we just discarded the additional map instead of merging.


radeksimko (approved):
Good catch 😅

PR #305, opened 2023-08-24
Remove execution of validation funcs

dbanck:
We move this to the language server in https://github.com/hashicorp/terraform-ls/pull/1361 to decouple it from schema validation and have more fine-grained control over when it runs.


radeksimko (approved):
https://github.com/hashicorp/hcl-lang/assets/287584/ebe861ad-695c-4c38-bb03-b43214205224

PR #310, opened 2023-08-31
Introduce `decoder.ValidateFile`

dbanck:
This PR adds a new method to the decoder for validating a single file

### DIFF
--- a/decoder/validate.go
+++ b/decoder/validate.go
@@ -48,7 +48,7 @@ func (d *PathDecoder) Validate(ctx context.Context) (lang.DiagnosticsMap, error)
 	return diags, nil
 }
 
-// ValidateFile returns a list of Diagnostics for a specific file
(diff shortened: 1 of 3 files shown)


decoder/validate.go
```diff
@@ -48,6 +48,28 @@ func (d *PathDecoder) Validate(ctx context.Context) (lang.DiagnosticsMap, error)
...
+	}
+
+	f, err := d.fileByName(filename)
+	if err != nil {
+		return hcl.Diagnostics{}, err
+	}
+
+	body, ok := f.Body.(*hclsyntax.Body)
+	if !ok {
+		// TODO! error
```
radeksimko, line 64: 👀 

I think we should mimic what we do in other similar methods already - i.e. return error to say that JSON isn't supported.

radeksimko (approved):
Dog Human Fashion

PR #315, opened 2023-09-06
go.mod: Fix `go` directive to include patch version

radeksimko:
This reflects upstream changes as described in https://github.com/golang/go/issues/62278 and unblocks dependabot which currently cannot run.

Screenshot 2023-09-06 at 07 21 07

An alternative fix would be to add `toolchain 1.21.0` but given that we don't need different version for compatibility and toolchain I think it's easier to just update the `go` directive for now.

See also https://github.com/dependabot/dependabot-core/issues/7895

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/hashicorp/hcl-lang
 
-go 1.21
+go 1.21.0
 
 require (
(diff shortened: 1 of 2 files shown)


dbanck (approved):
Nice catch

PR #318, opened 2023-09-19
decoder: Decouple validators & make them pluggable via PathContext

radeksimko:
This decoupling is aiming mainly to unblock support for multiple language IDs (schemas) with different needs for validation as shown in https://github.com/hashicorp/terraform-ls/pull/1413

It's expected to be useful outside of Terraform LS though in the future.


decoder/validate.go
```diff
@@ -74,6 +63,10 @@ type validationWalker struct {
 }
 
 func (vw validationWalker) Visit(ctx context.Context, node hclsyntax.Node, nodeSchema schema.Schema) (diags hcl.Diagnostics) {
+	if vw.validators == nil {
```
dbanck, line 66:
```suggestion
	if len(vw.validators) == 0 {
```

Do we want to add a similar check in `Validate()` to avoid iterating and casting all module files if there are no validators?

radeksimko: This was originally intended to be a `nil` check to prevent crashes but I only now realised that's not necessary as uninitialised slice is treated the same as empty one.

https://go.dev/play/p/KzpM-N1sGKF

I don't think we need any check to "avoid iterating" since there won't be any iterating for empty slices anyway, or the negligible benefit (possibly some low microseconds if even that) doesn't seem to be worth it, compared to the additional LOC.

dbanck: I was referring to the outer loop over the module files, before walking each file:
https://github.com/hashicorp/hcl-lang/pull/318/files#diff-943adc0589a4b4188c39ef5078910795369b4270122bb9a430e863a3a96ba666R25-R30

Of course, we can avoid the extra LOCs if the cost of casting each file body is negligible.

radeksimko: The casting shouldn't happen if there the slice is empty or uninitialised, the only time it could be "wasted" compute is when we have JSON files which we won't validate but I don't see how we could avoid _that_ situation as we have to range over it at least once to see the type, so any upfront filtering would involve casting anyway.

dbanck (approved):
👍

PR #322, opened 2023-09-27
decoder: Support template expressions

dbanck:
* Part of https://github.com/hashicorp/terraform-ls/issues/496
* Related to https://github.com/hashicorp/terraform-ls/issues/522

---

This PR adds support for template expressions inside single and multiline strings.

I filed https://github.com/hashicorp/hcl-lang/issues/345 to track edge cases for empty template expressions inside maps and objects.

### Directives

`if` directives within template expressions are [turned into conditional expressions](https://github.com/hashicorp/hcl/blob/c964a71ca32006c9e7a0730b5a0e1dd60b05d308/hclsyntax/parser_template.go#L242-L248) and `for` directives are [turned into for expressions](https://github.com/hashicorp/hcl/blob/c964a71ca32006c9e7a0730b5a0e1dd60b05d308/hclsyntax/parser_template.go#L331-L341). So I would defer support for this until those expressions land
* https://github.com/hashicorp/terraform-ls/issues/527 
* https://github.com/hashicorp/terraform-ls/issues/528

### DIFF
--- a/decoder/expr_any_hover.go
+++ b/decoder/expr_any_hover.go
@@ -96,7 +96,6 @@ func (a Any) HoverAtPos(ctx context.Context, pos hcl.Pos) *lang.HoverData {
 }
 
 func (a Any) hoverNonComplexExprAtPos(ctx context.Context, pos hcl.Pos) *lang.HoverData {
-	// TODO: Support TemplateExpr https://github.com/hashicorp/terraform-ls/issues/522
 	// TODO: Support splat expression https://github.com/hashicorp/terraform-ls/issues/526
 	// TODO: Support for-in-if expression https://github.com/hashicorp/terraform-ls/issues/527
 	// TODO: Support conditional expression https://github.com/hashicorp/terraform-ls/issues/528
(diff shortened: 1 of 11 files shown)


decoder/expr_any_hover.go
```diff
@@ -192,3 +195,34 @@ func (a Any) hoverOperatorExprAtPos(ctx context.Context, pos hcl.Pos) (*lang.Hov
...
+		if eType.IsStringLiteral() {
+			return nil, false
+		}
+
+		for _, partExpr := range eType.Parts {
+			if partExpr.Range().ContainsPos(pos) {
+				cons := schema.AnyExpression{
+					OfType: cty.String,
+				}
+				return newExpression(a.pathCtx, partExpr, cons).HoverAtPos(ctx, pos), true
```
radeksimko, line 211: It looks like the literal strings still need to be accounted for when they're inside of HEREDOC here, because we end up in https://github.com/hashicorp/hcl-lang/blob/0c7d286e56219c3b3c05a7124ee8547b1e710bc2/decoder/expr_literal_type_hover.go#L27-L42

decoder/expr_any_semtok.go
```diff
@@ -186,3 +188,34 @@ func (a Any) semanticTokensForOperatorExpr(ctx context.Context) ([]lang.Semantic
...
+
+		for _, partExpr := range eType.Parts {
+			cons := schema.AnyExpression{
+				OfType: cty.String,
+			}
+			expr := newExpression(a.pathCtx, partExpr, cons)
+			tokens = append(tokens, expr.SemanticTokens(ctx)...)
+		}
+
+		return tokens, true
```
radeksimko, line 209: I'm not sure why but it looks like that highlighting of HEREDOC still needs accounting for.

Screenshot 2023-11-21 at 13 09 43

radeksimko (approved):
LGTM aside from the support of HEREDOC in hover and semantic tokens.

I'm happy to finish it given your upcoming OOO though.

PR #330, opened 2023-10-06
(fix) decoder: Ensure empty objects & tuples still get collected as targets

radeksimko:
As reported in https://github.com/hashicorp/vscode-terraform/issues/1576 we previously did not collect empty objects and tuples as targets, which caused us to raise this as diagnostics that wasn't relevant.

This PR fixes it and adds some additional tests for some other types just to ensure that we don't introduce similar problem in the future there either.


jpogran (approved):
straightforward :shipit:

PR #333, opened 2023-10-11
decoder: allow objects in `for_each`

radeksimko:
Related to:

 - https://github.com/hashicorp/vscode-terraform/issues/1586

--- 

Relevant LOC in Terraform: https://github.com/hashicorp/terraform/blob/07b814c030daae39dea764571bef9cecf26e49be/internal/terraform/eval_for_each.go#L275-L284

As the in-line comment says:

Objects are still supported for backwards-compatible reasons but in general should be avoided. We add it here because otherwise valid configuration would be flagged as invalid. We cannot suppress it in completion (yet) or otherwise inform the user of this anti-pattern though.

## Before

Screenshot 2023-10-11 at 15 28 32

## After

Screenshot 2023-10-11 at 15 28 49

### DIFF
--- a/decoder/body_extensions_test.go
+++ b/decoder/body_extensions_test.go
@@ -667,7 +667,7 @@ func TestCompletionAtPos_BodySchema_Extensions_ForEach(t *testing.T) {
 					Value: "A meta-argument that accepts a map or a set of strings, and creates an instance for each item in that map or set.\n\n**Note**: A given block cannot use both `count` and `for_each`.",
 					Kind:  lang.MarkdownKind,
 				},
-				Detail: "optional, map of any single type or set of string",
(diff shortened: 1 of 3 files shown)


dbanck (approved):
LGTM! 

Too bad we can't suppress it on completion. Should we create a follow-up ticket for that?

radeksimko:
> Too bad we can't suppress it on completion. Should we create a follow-up ticket for that?

Yes - I think it should probably get suppressed in hover data as well. We could also consider separate explicit warning or info diagnostic to discourage users from using objects in that context.

I will create a ticket and track it on my TODOs in the meantime.

radeksimko:
Follow-up issue created: https://github.com/hashicorp/terraform-ls/issues/1473

PR #339, opened 2023-11-02
decoder: Ensure partially unknown dependent body is handled

radeksimko:
As demonstrated by the attached test, we can have a situation of "nested dependent modules" which was previously handled incorrectly. This is best explained on an example.

```hcl
variable "terraform_rs" {
  default = {
    backend              = "azurerm"
    workspace            = "***"
    storage_account_name = "***"
    container_name       = "***"
    key                  = "***"
    access_key           = "***"
  }
}

data "terraform_remote_state" "this" {
  backend   = var.terraform_rs.backend
  workspace = var.terraform_rs.workspace
  config = {
    storage_account_name = var.terraform_rs.storage_account_name
    container_name       = var.terraform_rs.container_name
    key                  = var.terraform_rs.key
    access_key           = var.terraform_rs.access_key
  }
}
```

Above, we have multiple dependent body schemas for the `data` block:

 - The first one is dependent on 1st label matching `terraform_remote_state` - that brings us `backend`, `workspace` and other attributes generally relevant for that particular data source
 - The second one is then dependent on the value of `backend` (i.e. attribute which was brought in as dependent from the earlier merging).

In the example above we would succeed in the first lookup, but fail in the second one - because we don't evaluate the expression in this context and cannot tell what `var.terraform_rs.backend` actually is. This is basically equivalent to any other invalid value, such as `backend = "foobar"` but the example above is a little more realistic. Now we report a more granular state reflecting the reality more accurately.

This is perhaps not ideal in the sense that although we suppress irrelevant diagnostics, we also suppress some helpful ones, e.g. invalid `backend`

```hcl
data "terraform_remote_state" "this" {
  backend   = "foobar"
  workspace = "noot"
}
```

Also we could run the validation in the top example with interpolated `var.terraform_rs.backend`, i.e. validate `config` in the context of `backend = "azurerm"` but interpolation is whole another problem we haven't really tackled yet.

We can treat this as an opportunity for future enhancements while reducing false negatives in the meantime.

## UX Before

Screenshot 2023-11-02 at 14 28 16

## UX After

Screenshot 2023-11-02 at 14 30 28

### DIFF
--- a/decoder/hover.go
+++ b/decoder/hover.go
@@ -163,8 +163,8 @@ func (d *PathDecoder) hoverContentForLabel(i int, block *hclsyntax.Block, bSchem
 	labelSchema := bSchema.Labels[i]
 
 	if labelSchema.IsDepKey {
-		bs, _, ok := schemahelper.NewBlockSchema(bSchema).DependentBodySchema(block.AsHCLBlock())
-		if ok {
+		bs, _, result := schemahelper.NewBlockSchema(bSchema).DependentBodySchema(block.AsHCLBlock())
+		if result == schemahelper.LookupSuccessful || result == schemahelper.LookupPartiallySuccessful {
 			content := fmt.Sprintf("`%s`", value)
 			if bs.Detail != "" {
 				content += " " + bs.Detail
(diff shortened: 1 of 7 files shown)


dbanck (approved):
👍 Appreciate the detailed examples

PR #340, opened 2023-11-03
Fix type-unaware target collection

dbanck:
While working on https://github.com/hashicorp/terraform-schema/pull/292, we discovered that we failed to collect type-unaware targets for attributes with `AsReference` enabled. 

This also led to the discovery that we were incorrectly creating type-aware targets with a NilType, making them a type-unaware target.

This PR fixes all of that, allowing the correct collection of local targets for https://github.com/hashicorp/terraform-schema/pull/292.


radeksimko (approved):
LGTM

PR #341, opened 2023-11-03
walker: Pass through context to enable more granular validation

radeksimko:
See https://github.com/hashicorp/terraform-ls/pull/1485


decoder/internal/walker/walker.go
```diff
@@ -24,14 +24,23 @@ type Walker interface {
...
+		blkNestingLvl = 0
+	}
+
 	switch nodeType := node.(type) {
 	case *hclsyntax.Body:
+		bodyCtx := ctx
 		foundBlocks := make(map[string]uint64)
 		dynamicBlocks := make(map[string]uint64)
-		bodySchema, ok := nodeSchema.(*schema.BodySchema)
-		if ok {
```
dbanck, line 32: Why are we getting rid of this check here? I see you moved it to the two for loops for attribute and blocks.
With this change, we call walk in lines 60 and 87, even without schemas.

radeksimko: This was done in a separate commit - which I hoped would provide the context/clarity 😅 https://github.com/hashicorp/hcl-lang/pull/341/commits/6d24e31a06cd05f1c058dcf1d7c38ac82798467b

The point is that we previously skipped walking body for which we don't have schema. Now this is _probably_ fine for all the existing use cases we have for the validation as we cannot really do much validation without schema anyway and all the validators short-circuit on lack of schema early on. However, this is inconsistent with how we treat other AST nodes (blocks or attributes), so I made it consistent here.

It also helped in the walker tests where I did not have to supply any schema just to verify that walking works.

dbanck: I see, that makes sense. But do we want to report diagnostics for every attribute/block within a block without a schema? Feels nosy to me.

**Before**
CleanShot 2023-11-07 at 13 26 51@2x

**After**
CleanShot 2023-11-07 at 13 24 24@2x

radeksimko: Good question. I _think_ we could suppress that mentioned noise by adding `ctx = schemacontext.WithUnknownSchema(ctx)` - equivalent to when merging fails?

I can add that patch.

radeksimko: done, PTAL

dbanck (approved):
👍 thanks for addressing this

Issue #280, opened 2023-05-24
Broken reference completion w/ trailing dot inside complex types `{}` / `[]`


radeksimko:
## Context

Reference completion often includes trailing dot, which is the step separator. The trailing dot is not included in the AST by HCL, which we already account for in the obvious cases:

https://github.com/hashicorp/hcl-lang/blob/bd545f4d6346360fb9a4e2e4a81423fa212c581e/decoder/expr_reference_completion.go#L71-L82

However we do not seem to account for similar cases involving references with trailing dot inside complex types (list, set, map, maybe object as well):

![2023-05-24 10 15 40](https://github.com/hashicorp/hcl-lang/assets/287584/5830bf08-6a95-4696-a752-7a00daf85e65)

## Debugging Details

For list specifically I was able to narrow down the codepath to this condition, which compares the element range with the given position, which is _not_ contained (it's off by one due to the dot that isn't included in the range):

https://github.com/hashicorp/hcl-lang/blob/bd545f4d6346360fb9a4e2e4a81423fa212c581e/decoder/expr_list_completion.go#L77-L79

It is likely that other complex types have the same root cause.

## Proposal

Fix the bug by providing completion of nested references after the trailing dot.

dbanck:
Here is a test case for objects

```go

func TestCompletionAtPos_exprObject_references(t *testing.T) {
	testCases := []struct {
		testName           string
		attrSchema         map[string]*schema.AttributeSchema
		refTargets         reference.Targets
		cfg                string
		pos                hcl.Pos
		expectedCandidates lang.Candidates
	}{
		{
			"single-line element with trailing dot",
			map[string]*schema.AttributeSchema{
				"attr": {
					Constraint: schema.Object{
						Attributes: schema.ObjectAttributes{
							"foo": {
								IsOptional: true,
								Constraint: schema.Reference{OfScopeId: lang.ScopeId("variable")},
							},
						},
					},
				},
			},
			reference.Targets{
				{
					Addr: lang.Address{
						lang.RootStep{Name: "var"},
						lang.AttrStep{Name: "bar"},
					},
					RangePtr: &hcl.Range{
						Filename: "variables.tf",
						Start:    hcl.Pos{Line: 2, Column: 1, Byte: 17},
						End:      hcl.Pos{Line: 2, Column: 3, Byte: 19},
					},
					ScopeId: lang.ScopeId("variable"),
				},
			},
			`attr = { foo = var. }
`,
[...]

PR #344, opened 2023-11-13
Improve traversal completion for lists, tuples, and sets

dbanck:
This PR improves the partial traversal completion within lists, tuples, and sets by recovering the trailing dot.

Since I recently worked on this for template expressions, I thought I could fix it for lists, tuples, and sets as well.

### UX After

Closes https://github.com/hashicorp/vscode-terraform/issues/1624


radeksimko (approved):
LGTM, small nitpick: I'd replace the `OneOf` with the actual type in tests since it should have no impact on what's being tested.

Also if we can either address or ticket out the other types like tuple, object and map (or confirm they're not impacted) that would be great.

dbanck:
I fixed the issues for tuples and found out that maps and objects are also affected. But we already track these in https://github.com/hashicorp/hcl-lang/issues/280

PR #353, opened 2023-12-21
decoder: Add new `LanguageIDs` method to `PathReader`

dbanck:
WIP

### DIFF
--- a/decoder/path_reader.go
+++ b/decoder/path_reader.go
@@ -11,7 +11,8 @@ import (
 )
 
 type PathReader interface {
-	Paths(ctx context.Context) []lang.Path
+	Paths(ctx context.Context, languageID string) []lang.Path
+	LanguageIDs() []string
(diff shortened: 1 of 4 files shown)


decoder/path_reader.go
```diff
@@ -11,7 +11,8 @@ import (
 )
 
 type PathReader interface {
-	Paths(ctx context.Context) []lang.Path
+	Paths(ctx context.Context, languageID string) []lang.Path
+	LanguageIDs() []string
```
radeksimko, line 15:
```suggestion
	LanguageIDs() []string
	Paths(ctx context.Context, languageID string) []lang.Path
```
Nitpick - this just reflects the order in which the methods get called.

dbanck:
Superseded by a another approach in https://github.com/hashicorp/terraform-ls/pull/1667

Issue #267, opened 2023-04-25
decoder/schema: Remove old (unused) logic related to legacy expressions


radeksimko:
## Context

As part of some recent PRs we introduced `Constraint` to `schema.AttributeSchema`:

https://github.com/hashicorp/hcl-lang/blob/9c3a246c9754c6dda052c6874680023624ec46ee/schema/attribute_schema.go#L26-L28

This is now used as the primary way to describe attribute's RHS (expression) constraints.

This leaves us with the old/legacy way still in place which is now unused

https://github.com/hashicorp/hcl-lang/blob/9c3a246c9754c6dda052c6874680023624ec46ee/schema/attribute_schema.go#L22-L24

and also no longer covered by tests https://github.com/hashicorp/hcl-lang/pull/251

## Proposal

 - [ ] schema: Remove `Expr` from `AttributeSchema`
 - [ ] schema: Remove any logic related to `Expr`
 - [ ] decoder: Remove any logic related to `Expr`

PR #354, opened 2023-12-22
Remove old (unused) logic related to legacy expressions

dbanck:
Closes https://github.com/hashicorp/hcl-lang/issues/267

### TODO

- [ ] Do we want to keep the legacy tests which have been updated to use the new `Constraint` in #251


radeksimko (approved):
giphy

--- 

> Do we want to keep the legacy tests which have been updated to use the new Constraint?

I think the simple answer here would be: If they cover the same LOC (as per `go test -cover`) or we at least achieve the same % coverage as without them, then we should feel free to remove them. If not then we should investigate the delta and maybe keep those tests and/or rewrite them.

In the interest of keeping PRs small and opened for short time though I'd leave this for another one.

PR #355, opened 2023-12-22
Extract smaller expressions into separate files

dbanck:
This PR extracts the template expressions and operator expressions into separate files as @jpogran suggested.

We already did this for the conditional expressions.


radeksimko (approved):
LGTM, good idea.

I am hoping that one day we can also separate out all the expression related logic into its own package as there's a _lot_ of files now concerning just that, mixed up with everything else.

PR #358, opened 2024-01-05
decoder: remove dead code

giautm:


radeksimko (approved):
Good catch! Thank you for the PR.

We should probably plug some "linter" into the pipeline which detects dead code but we haven't had the time/chance to investigate options and choose (and repeat that for every relevant repo the team manages yet).

PR #360, opened 2024-01-08
Remove legacy tests

dbanck:
This PR deletes the converted legacy tests for the recently deleted expressions in #354.

### Coverage

```
Pre removal
ok      github.com/hashicorp/hcl-lang/decoder   2.683s  coverage: 84.8% of statements


After removal
ok      github.com/hashicorp/hcl-lang/decoder   2.385s  coverage: 83.7% of statements


Keeping the hook tests
ok      github.com/hashicorp/hcl-lang/decoder   2.186s  coverage: 84.6% of statements
```

We lose a little bit of coverage in some functions, but only about 0.2% overall. I can look into that more if you think it's important to cover those branches.

```
expr_literal_type_semtok.go:16:           SemanticTokens                          91.4% --> 89.7%
expr_literal_value_hover.go:17:           HoverAtPos                              83.2% --> 82.2%
expr_literal_value_semtok.go:15:          SemanticTokens                          81.3% --> 80.4%           
expr_object_completion.go:21:             CompletionAtPos                         90.3% --> 88.9%
expr_object_hover.go:16:                  HoverAtPos                              96.0% --> 92.0%
expr_reference_completion.go:15:          CompletionAtPos                         97.1% --> 94.3%
expression.go:272:                        rawObjectKey                            82.6% --> 78.3%
expression_candidates.go:392:             labelForLiteralValue                    92.0% --> 90.0%
semantic_tokens.go:132:                   isPrimitiveTypeDeclaration              85.7% --> 71.4%
```

I can send you the html coverage report, if you want to dig deeper.


decoder/expression_candidates_test.go
```diff
@@ -0,0 +1,248 @@
+package decoder
```
radeksimko, line 1: Looks like this file is missing the license header

radeksimko (commented):
> I can send you the html coverage report, if you want to dig deeper.

We can always generate it later if we need as the code will be in git history, so don't worry about it.

radeksimko (approved):

PR #361, opened 2024-01-08
Remove unused expression helpers

dbanck:
In #354, I missed some helper functions for the old expression logic


radeksimko (approved):
Good catch, it would be nice to have some dead code detector in the pipeline some day 😅

PR #362, opened 2024-01-09
Rename `*Traversal*` to `*Reference*`

### LINKED ISSUE #266: schema/lang: Rename `Traversal*` to `Reference*`
## Depends on

 - https://github.com/hashicorp/hcl-lang/issues/267

## Context

As part of https://github.com/hashicorp/hcl-lang/pull/185 and some other related PRs we started using a slightly different naming convention (traversal -> reference essentially).

We used the term "reference" in other places before (there's whole package `reference`), so it makes sense to align the rest of the codebase and get rid of "traversal", which is usually only how we refer to references in HCL internals. The aim is to make the public API friendlier to developers who have to build the schema.

Screenshot 2023-04-25 at 14 05 16

## Proposal

 - [ ] Rename all existing types/functions which involve "traversal" to "reference".
 - [ ] Update any references in `terraform-schema`
 - [ ] Update any references in `terraform-ls`

dbanck:
Closes #266

### DIFF
--- a/lang/semantic_token.go
+++ b/lang/semantic_token.go
@@ -30,7 +30,7 @@ const (
 	TokenObjectKey     SemanticTokenType = "hcl-objectKey"
 	TokenMapKey        SemanticTokenType = "hcl-mapKey"
 	TokenKeyword       SemanticTokenType = "hcl-keyword"
-	TokenTraversalStep SemanticTokenType = "hcl-traversalStep"
+	TokenReferenceStep SemanticTokenType = "hcl-referenceStep"
 	TokenTypeComplex   SemanticTokenType = "hcl-typeComplex"
 	TokenTypePrimitive SemanticTokenType = "hcl-typePrimitive"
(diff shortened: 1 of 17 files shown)


lang/semantic_token.go
```diff
@@ -30,7 +30,7 @@ const (
 	TokenObjectKey     SemanticTokenType = "hcl-objectKey"
 	TokenMapKey        SemanticTokenType = "hcl-mapKey"
 	TokenKeyword       SemanticTokenType = "hcl-keyword"
-	TokenTraversalStep SemanticTokenType = "hcl-traversalStep"
+	TokenReferenceStep SemanticTokenType = "hcl-referenceStep"
```
radeksimko, line 33: I think we'll also need to reflect this in both LS (incl. docs) and vscode-terraform (incl. docs).

radeksimko (approved):
It would be useful to see downstream PRs in `terraform-ls` and `vscode-terraform` passing before we merge this but otherwise LGTM.

PR #366, opened 2024-01-16
decoder: Add support for parenthesis on RHS

radeksimko:
Closes https://github.com/hashicorp/terraform-ls/issues/529

Operators were already implemented as part of https://github.com/hashicorp/hcl-lang/pull/320

--- 

This enables support for parenthesis on the RHS (right hand side), i.e. after `=`.

For example:

```hcl
attr = (var.foo + var.bar) * 2
```

### Completion

Screenshot 2024-01-18 at 20 43 49

Screenshot 2024-01-18 at 20 43 31

### Hover

Screenshot 2024-01-18 at 20 44 19

### Semantic Tokens

Screenshot 2024-01-18 at 20 44 41

### Reference Origins

--- 

## Follow-up Issues

 - https://github.com/hashicorp/terraform-ls/issues/1586

### DIFF
--- a/decoder/expr_any_completion_test.go
+++ b/decoder/expr_any_completion_test.go
@@ -3426,6 +3426,219 @@ func TestCompletionAtPos_exprAny_operators(t *testing.T) {
 	}
 }
 
+func TestCompletionAtPos_exprAny_parentheses(t *testing.T) {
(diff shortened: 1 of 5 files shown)


dbanck (approved):
Nice!

PR #367, opened 2024-01-18
decoder: Add support for parenthesis on LHS (map keys & attribute names)

radeksimko:
Depends on:

 - https://github.com/hashicorp/hcl-lang/pull/366

--- 

This enables support for interpolated map keys and object attribute names,. Since this is enabled by parenthesis it is closely related to that linked PR which adds support for that expression type.

In other words this enables parenthesis support on the LHS (left hand side).

Example:

```hcl
map_attr = {
  (var.keyname) = "foo"
}
```

## Implementation Notes

This does _not_ enable equivalent support inside of inferred maps/objects, such as in the context of Terraform's `locals`. This is currently tracked under https://github.com/hashicorp/terraform-ls/issues/1586 

It originally started out as "hidden implementation detail" of `decoder` (the decision whether interpolation is allowed was originally just implied by `AnyExpression{}` until I realised that we still use `schema.Map{}` and `schema.Object{}` in contexts where we expect interpolation, including interpolated keys, so I had to walk it back and make it a more 1st class feature of the schema, for better or worse.

For that reason, this will also require some changes downstream:

 - https://github.com/hashicorp/terraform-schema/pull/316

### Completion

Screenshot 2024-01-18 at 20 47 21

### Hover

Screenshot 2024-01-18 at 20 49 08

### Semantic Tokens

Screenshot 2024-01-18 at 20 49 28

### DIFF
--- a/decoder/expr_object_semtok.go
+++ b/decoder/expr_object_semtok.go
@@ -52,12 +52,9 @@ func (obj Object) SemanticTokens(ctx context.Context) []lang.SemanticToken {
 				// TODO: Consider not reporting the quotes?
 				Range: item.KeyExpr.Range(),
 			})
-		}
 
-		if isKnownAttr {
 			expr := newExpression(obj.pathCtx, item.ValueExpr, aSchema.Constraint)
 			tokens = append(tokens, expr.SemanticTokens(ctx)...)
-			continue
 		}
 	}
 
--- a/decoder/expr_object_semtok.go
+++ b/decoder/expr_object_semtok.go
@@ -7,7 +7,9 @@ import (
 	"context"
 
 	"github.com/hashicorp/hcl-lang/lang"
+	"github.com/hashicorp/hcl-lang/schema"
 	"github.com/hashicorp/hcl/v2/hclsyntax"
+	"github.com/zclconf/go-cty/cty"
 )
 
 func (obj Object) SemanticTokens(ctx context.Context) []lang.SemanticToken {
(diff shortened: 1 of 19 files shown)


schema/constraint_object.go
```diff
@@ -25,6 +25,10 @@ type Object struct {
 
 	// Description defines description of the whole object (affects hover)
 	Description lang.MarkupContent
+
+	// AllowInterpolatedAttrName determines whether the attribute names can be
+	// interpolated (true) or static (literal strings only).
+	AllowInterpolatedAttrName bool
```
dbanck, line 31: While `AllowInterpolatedAttrName` is technically more correct, I would suggest using `AllowInterpolatedKeys` as the field name here. This makes it easier to read and understand, and we have a single field/concept to enable the same behavior for maps and objects. In other places, like the comment in line 19 or `rawObjectKeys`, we also use the term `keys` for objects.

radeksimko: Strictly speaking, the key becomes attribute when it is treated as an object. Perhaps confusingly (yet intentionally), the reason `rawObjectKeys` uses the term _key_ is because the context is `hclsyntax.ObjectConsExpr`. To put it differently, HCL's term for `{}` is "object constant" whether it ends up being a map or an object. The possibly confusing part is that HCL uses terms as "key" and "value" (in spite of referring to it as object), so this was all mostly in alignment with upstream naming.

All that said, I am not opposed to renaming this to `AllowInterpolatedKeys`, while not strictly "correct" it's not also entirely incorrect and it's pragmatic.

I can commit that patch.

decoder/expr_object_semtok.go
```diff
@@ -23,27 +25,40 @@ func (obj Object) SemanticTokens(ctx context.Context) []lang.SemanticToken {
...
+				Type:      lang.TokenObjectKey,
+				Modifiers: lang.SemanticTokenModifiers{},
+				// TODO: Consider not reporting the quotes?
+				Range: item.KeyExpr.Range(),
+			})
+		}
 
-		expr := newExpression(obj.pathCtx, item.ValueExpr, aSchema.Constraint)
-		tokens = append(tokens, expr.SemanticTokens(ctx)...)
+		if isKnownAttr {
```
dbanck, line 57: We can combine this with the `if` block above

decoder/expr_object_semtok.go
```diff
@@ -23,27 +25,40 @@ func (obj Object) SemanticTokens(ctx context.Context) []lang.SemanticToken {
...
+			expr := newExpression(obj.pathCtx, item.ValueExpr, aSchema.Constraint)
+			tokens = append(tokens, expr.SemanticTokens(ctx)...)
+			continue
```
dbanck, line 60:
```suggestion
```

No need for `continue` at the end of the loop

dbanck (approved):
LGTM! Just two small nits

PR #368, opened 2024-01-22
Support `for`-`in`-`if` expressions

radeksimko:
Closes https://github.com/hashicorp/terraform-ls/issues/527

--- 

## Follow-up Issues

 - [ ] https://github.com/hashicorp/terraform-ls/issues/1607
 - [ ] https://github.com/hashicorp/terraform-ls/issues/1608
 - [ ] https://github.com/hashicorp/terraform-ls/issues/1609
 - [ ] https://github.com/hashicorp/hcl-lang/issues/370

### DIFF
--- a/decoder/expr_any_completion_test.go
+++ b/decoder/expr_any_completion_test.go
@@ -4000,6 +4000,802 @@ func TestCompletionAtPos_exprAny_parentheses(t *testing.T) {
 	}
 }
 
+func TestCompletionAtPos_exprAny_forExpr(t *testing.T) {
+	testCases := []struct {
+		testName           string
+		attrSchema         map[string]*schema.AttributeSchema
+		refTargets         reference.Targets
+		cfg                string
+		pos                hcl.Pos
+		expectedCandidates lang.Candidates
+	}{
+		// list
+		{
+			"list on collection",
+			map[string]*schema.AttributeSchema{
+				"attr": {
+					Constraint: schema.AnyExpression{
+						OfType: cty.List(cty.String),
+					},
+				},
+			},
+			reference.Targets{
+				{
+					Addr: lang.Address{
+						lang.RootStep{Name: "var"},
+						lang.AttrStep{Name: "foo"},
+					},
+					Type: cty.List(cty.String),
+				},
+				{
+					Addr: lang.Address{
+						lang.RootStep{Name: "var"},
+						lang.AttrStep{Name: "mystring"},
+					},
+					Type: cty.String,
+				},
+			},
+			`attr = [for i, v in var: v]
+`,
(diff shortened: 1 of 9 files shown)


decoder/expr_any_completion_test.go
```diff
@@ -4000,6 +4000,802 @@ func TestCompletionAtPos_exprAny_parentheses(t *testing.T) {
...
+				},
+				{
+					Addr: lang.Address{
+						lang.RootStep{Name: "var"},
+						lang.AttrStep{Name: "bar"},
+					},
+					Type: cty.String,
+				},
+			},
+			`attr = [for i, v in var: v]
```
dbanck, line 4083: I found this confusing at first, because I was expecting mainly `v` to be used here. But after giving it some thought, it makes sense to have the full completion list here.

decoder/expr_any_for.go
```diff
@@ -0,0 +1,289 @@
...
+	"github.com/hashicorp/hcl/v2"
+	"github.com/hashicorp/hcl/v2/hclsyntax"
+	"github.com/zclconf/go-cty/cty"
+)
+
+func (a Any) completeForExprAtPos(ctx context.Context, pos hcl.Pos) ([]lang.Candidate, bool) {
+	candidates := make([]lang.Candidate, 0)
+
+	switch eType := a.expr.(type) {
+	case *hclsyntax.ForExpr:
```
dbanck, line 21: Can we recover the `.` when triggering completion with a prefix?

CleanShot 2024-02-06 at 10 24 57@2x

dbanck (approved):
Nice work!

Just a question/suggestion that could also be addressed in a follow-up issue.

decoder/expr_any_completion_test.go
radeksimko: The `v` as in the value reference support is something now tracked under https://github.com/hashicorp/terraform-ls/issues/1608

decoder/expr_any_for.go
radeksimko: That would be a little more difficult here unfortunately as we'd have to be recovering pretty much everything up to the `[` _at least_. I didn't even think of the consequences of the incomplete `ForExpr` wrapped into another expression like function here but I expect it to be similarly difficult.

Either way we now track it under https://github.com/hashicorp/hcl-lang/issues/370

radeksimko:
@dbanck I no longer have push/merge permissions here, so I'll have to leave it up to you to merge 😅

lukaabicic:
Please fix variable declaration not found error in [2.29.4] version (2024-02-13). 
The error occurs when the resources are created with a for_each argument whose values are variables of type list(object()).
The error isn't present in the previous versions.

PR #372, opened 2024-02-08
feat: support namespaced functions

ansgarm:
🚫 blocked until there's a new release of `hcl/release` with support for `ExprSyntaxError`:
https://github.com/hashicorp/hcl-lang/blob/905e5636294807904a67f6c89a0562db0b82f32c/go.mod#L8

Closes https://github.com/hashicorp/vscode-terraform/issues/1697

### DIFF
--- a/decoder/expr_function.go
+++ b/decoder/expr_function.go
@@ -151,9 +151,8 @@ func (fe functionExpr) HoverAtPos(ctx context.Context, pos hcl.Pos) *lang.HoverD
 
 	if funcExpr.NameRange.ContainsPos(pos) {
 		return &lang.HoverData{
-			Content: lang.Markdown(fmt.Sprintf("```terraform\n%s(%s) %s\n```\n\n%s",
-				funcExpr.Name, parameterNamesAsString(funcSig), funcSig.ReturnType.FriendlyName(), funcSig.Description)),
-			Range: fe.expr.Range(),
+			Content: hoverContentForFunction(funcSig, funcExpr),
+			Range:   fe.expr.Range(),
 		}
 	}
 
@@ -296,4 +295,9 @@ func (fe functionExpr) matchingFunctions(prefix string, editRange hcl.Range) []l
 	})
 
 	return candidates
+}
+
+func hoverContentForFunction(funcSig schema.FunctionSignature, funcExpr *hclsyntax.FunctionCallExpr) lang.MarkupContent {
+	return lang.Markdown(fmt.Sprintf("```terraform\n%s(%s) %s\n```\n\n%s\n\n%s",
+		funcExpr.Name, parameterNamesAsString(funcSig), funcSig.ReturnType.FriendlyName(), funcSig.Description, funcSig.Detail))
 }
--- a/decoder/expr_function.go
+++ b/decoder/expr_function.go
@@ -58,9 +58,7 @@ func (fe functionExpr) CompletionAtPos(ctx context.Context, pos hcl.Pos) []lang.
 		return fe.matchingFunctions(prefix, eType.Range())
 
 	case *hclsyntax.ExprSyntaxError:
-		// TODO: "provider:" does not jump here, seems to be some other expression?
-
-		// This range can range up until the end of the file in case of invalid config
+		// Note: this range can range up until the end of the file in case of invalid config
 		if eType.SrcRange.ContainsPos(pos) {
 			// we are somewhere in the range for this attribute but we don't have an expression range to check
 			// so we look back to check whether we are in a partially written provider defined function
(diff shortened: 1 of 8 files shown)


schema/signature.go
```diff
@@ -25,3 +27,15 @@ type FunctionSignature struct {
...
+
+func (fs *FunctionSignature) Copy() *FunctionSignature {
+	newFS := &FunctionSignature{
+		Description: fs.Description,
+		Detail:      fs.Detail,
+		ReturnType:  fs.ReturnType, // TODO: deep copy needed?
+		VarParam:    fs.VarParam,   // TODO: deep copy needed?
+	}
+	newFS.Params = make([]function.Parameter, len(fs.Params))
+	copy(newFS.Params, fs.Params) // TODO: deep copy needed?
```
ansgarm, line 39: @dbanck, do you think this needs a deep(er) copy?

dbanck: Yes, we usually need to make sure to explicitly copy slicing and pointing.

schema/signature.go
```diff
@@ -25,3 +27,34 @@ type FunctionSignature struct {
 	// parameter if it is supported.
 	VarParam *function.Parameter
 }
+
+func (fs *FunctionSignature) Copy() *FunctionSignature {
+	newFS := &FunctionSignature{
+		Description: fs.Description,
+		Detail:      fs.Detail,
+		ReturnType:  fs.ReturnType,
+		VarParam:    copyFunctionParameter(fs.VarParam),
```
dbanck, line 36: Instead of using a helper function here, that reimplements some of the `cty` internals (field names) and would need to be updated whenever something upstream changes, we can make a copy using some pointer magic.

```go
if fs.VarParam != nil {
	vpCpy := *fs.VarParam
	newFS.VarParam = &vpCpy
}
```

ansgarm: Ah, neat! And that works because there are no nested pointers, right? Or would that be irrelevant anyway?

dbanck: While it possible to have nested pointers in Go (`**T`), I don't think it's relevant here.

schema/signature.go
```diff
@@ -25,3 +27,20 @@ type FunctionSignature struct {
...
 }
+
+func (fs *FunctionSignature) Copy() *FunctionSignature {
+	newFS := &FunctionSignature{
+		Description: fs.Description,
+		Detail:      fs.Detail,
+		ReturnType:  fs.ReturnType,
+	}
+
+	copy(newFS.Params, fs.Params)
```
dbanck, line 38:
```suggestion
        newFS.Params = make([]function.Parameter, len(fs.Params))
	copy(newFS.Params, fs.Params)
```

We still need to create the slice before copying something into it.

decoder/expr_function.go
```diff
@@ -297,3 +296,14 @@ func (fe functionExpr) matchingFunctions(prefix string, editRange hcl.Range) []l
 
 	return candidates
 }
+
+func hoverContentForFunction(funcSig schema.FunctionSignature, funcExpr *hclsyntax.FunctionCallExpr) lang.MarkupContent {
```
dbanck, line 300:
```suggestion
func hoverContentForFunction(name string, funcSig schema.FunctionSignature) lang.MarkupContent {
```

Since we are using only one field (`name)`, we could pass it to the function as a string instead of passing the whole expression.

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.21.0
 require (
 	github.com/google/go-cmp v0.6.0
 	github.com/hashicorp/go-multierror v1.1.1
-	github.com/hashicorp/hcl/v2 v2.19.1
+	github.com/hashicorp/hcl/v2 v2.19.2-0.20240209134327-1e1a6b858a91
```
dbanck, line 8: I think we should wait for an HCL release before merging this

dbanck (changes requested):
Nice work, some minor suggestions.

Thanks for adding a namespaced function to the tests

schema/signature.go
ansgarm: Well that makes sense 😅

decoder/expr_function.go
```diff
@@ -55,6 +56,61 @@ func (fe functionExpr) CompletionAtPos(ctx context.Context, pos hcl.Pos) []lang.
 
 		prefix := rootName[0:prefixLen]
 		return fe.matchingFunctions(prefix, eType.Range())
+
+	case *hclsyntax.ExprSyntaxError:
+		// TODO: "provider:" does not jump here, seems to be some other expression?
```
ansgarm, line 61: Is this too edge-casey? Do we need a follow-up issue for this?

decoder/expr_function.go
```diff
@@ -55,6 +56,61 @@ func (fe functionExpr) CompletionAtPos(ctx context.Context, pos hcl.Pos) []lang.
...
+			recoveredSuffixBytes := recoverRightBytes(fileBytes, pos, func(offset int, r rune) bool {
+				return !isNamespacedFunctionNameRune(r) && r != '('
+			})
+			// recoveredSuffixBytes also contains the rune after the function name, so we need to trim it
+			_, lengthLastRune := utf8.DecodeLastRune(recoveredSuffixBytes)
+			recoveredSuffixBytes = recoveredSuffixBytes[:len(recoveredSuffixBytes)-lengthLastRune]
+
+			recoveredIdentifier := append(recoveredPrefixBytes, recoveredSuffixBytes...)
+
+			// TODO: this is specific to Terraform provider defined functions, we should generalize this
```
ansgarm, line 85: We are currently only looking for the `provider:` prefix, but in theory `hcl-lang` should support any namespaced function, right?

dbanck: Yes, we should avoid implementing anything Terraform specific in `hcl-lang`. Any namespaced function should work and also the short syntax `namespace::func()`

decoder/expr_function.go
dbanck: If you want, you could do a quick check to see if you can catch the `ScopeTraversalExpr` case, and if it gets too complicated, create a follow-on issue. But I'm fine with creating a follow-on issue right away.

ansgarm: Created a follow-up issue: https://github.com/hashicorp/hcl-lang/issues/385

dbanck (approved):

PR #381, opened 2024-03-01
fix panic for partial provider defined functions in tuple expressions (`[ provider:: ]`)

ansgarm:


ansgarm:
made obsolete by https://github.com/hashicorp/hcl/pull/668

PR #394, opened 2024-06-02
decoder: refactor `SelfRefs` use context

giautm:
This PR refactors a mess in the codebase with `allowSelfRefs` passing around until it reaches the place it was used for.

The helper `schema.WithActiveSelfRefs(ctx)` is already defined by the author, so I just use it for this suppose; and remove the `allowSelfRefs` variable from the method.

Before:
```go
type ReferenceOriginsExpression interface {
	ReferenceOrigins(ctx context.Context, allowSelfRefs bool) reference.Origins
}
```

After:
```go
type ReferenceOriginsExpression interface {
	ReferenceOrigins(ctx context.Context) reference.Origins
}
```

### DIFF
--- a/decoder/expr_any_conditional.go
+++ b/decoder/expr_any_conditional.go
@@ -70,7 +70,7 @@ func (a Any) hoverConditionalExprAtPos(ctx context.Context, pos hcl.Pos) (*lang.
 	return nil, false
 }
 
-func (a Any) refOriginsForConditionalExpr(ctx context.Context, allowSelfRefs bool) (reference.Origins, bool) {
+func (a Any) refOriginsForConditionalExpr(ctx context.Context) (reference.Origins, bool) {
 	origins := make(reference.Origins, 0)
 
 	// There is currently no way of decoding conditional expressions in JSON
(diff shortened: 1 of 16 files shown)


giautm:
@dbanck: please help to review

jpogran:
Hi @giautm, thank you for offering a contribution to this repo. Can you please put a description of your change in the PR, along with the reasoning for your change? Right now the PR description is empty and the commit message just points to https://github.com/giautm/hcl-lang/pull/2, which I did not find helpful either.

In general it is a good practice to open an issue describing the change you want so there is a chance for us to discuss the design before work is put into a PR that may not be accepted as is. Apologies, this repo does not have a CONTRIBUTING.md to have set that expectation, I will fix that.

Once we have an understanding of the intent of the change, we can review this PR.

giautm:
@jpogran updated its description, thank you.

giautm:
Hello @dbanck, any chance for a review?

ansgarm (approved):
Hi @giautm 👋

First of all: Thank you for this improvement and for taking care in shepherding this through 💯

I rebased your branch which allowed me to locally test whether your changes work with the latest `terraform-ls`, which they (unsurprisingly) do 🎉
I also added a test case that tests the code path you touched as I couldn't find any existing test that covered this for origins.

From my point of view this looks fine to merge, but I'll ask a colleague to have a quick look at the test I added and then we should be all set to merge this 👍 Thanks again for taking the time to improve this!

giautm:
Thank you for kind of words. If you don't mind, I would like to send more PRs later. I currently also working on another ls that also using hcl-lang, I found place the hcl-lang can be improve.

jpogran (approved):

PR #399, opened 2024-06-21
Add contributing guidelines

jpogran:
This copies our terraform-ls contributing guidelines and adjusts where appropriate. It is minimal right now, more sections can be added as we move towards a stable API


dbanck (approved):
Nice! Thanks for adding this

PR #404, opened 2024-07-18
feat: allow overriding existing blocks and attributes for dependent body schemas

ansgarm:
We couldn't find any cases where the existing tests (and some example codebase) would access the else statement while testing. However, this change could allow e.g. Terraform providers to override for example the lifecycle block if they define such a block in their schema. We have yet to see such cases though :)

### Open
- [x] Add a test case for this new behaviour (since the old didn't have one)

### DIFF
--- a/decoder/internal/schemahelper/block_schema.go
+++ b/decoder/internal/schemahelper/block_schema.go
@@ -29,29 +29,18 @@ func MergeBlockBodySchemas(block *hcl.Block, blockSchema *schema.BlockSchema) (*
 	depSchema, _, result := NewBlockSchema(blockSchema).DependentBodySchema(block)
 	if result == LookupSuccessful || result == LookupPartiallySuccessful {
 		for name, attr := range depSchema.Attributes {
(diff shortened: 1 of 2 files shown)


ansgarm:
Thanks for the quick reply!

Yeah, we were running into this while implementing support for provider blocks in stacks projects. Specifically, we have a static schema with an empty, default `config` block which can't be overridden via a dependent body schema if it's already defined.

```hcl
provider "aws" "this" {
  config {
    region = var.region

    assume_role_with_web_identity {
      role_arn                = var.role_arn
      web_identity_token_file = var.identity_token_file
    }
  }
}
```

radeksimko (approved):
I see, that is some unfortunate complexity that is being introduced 😑 but in that case the change makes sense!

I recall that multiple dependent bodies already made things complex enough when we were implementing the Terraform language alone. It's not easy to design a static schema on top of such dynamic behaviours.

dbanck (approved):

PR #407, opened 2024-08-07
Support OriginForTarget

ansgarm:


jpogran:
Completed in https://github.com/hashicorp/hcl-lang/pull/408

PR #410, opened 2024-08-21
(TF-19544) fix: Support reference completions with a trailing dot in object constructors

ansgarm:
This depends on https://github.com/hashicorp/hcl/pull/692

- **fix: Support reference completions after dot in object constructors**
- **Bump hcl to `117baa8`**

completions


jpogran (commented):
Might this go to main as it's not specific to stacks?

ansgarm:
> Might this go to main as it's not specific to stacks?

Good call! Yeah 👍 I'll rebase

jpogran (approved):
❤️

PR #415, opened 2024-10-07
Fix whitespace via `go fmt`

dbanck:
CI currently exits early on `main`, because the `go fmt` step fails


ansgarm (approved):
If it fails on `main`, maybe we want it to fail on PRs as well? 😄

dbanck:
👍 I will investigate

PR #442, opened 2025-01-17
feat: Collect write only attributes via PathDecoder

ansgarm:
This adds support for the `PathDecoder` to collect write only attributes used in resources


jpogran (approved):
❤️

PR #490, opened 2026-07-08
Propagate AnyAttribute when merging dependent block body schemas

akunaatrium:
## Summary

- Propagate `AnyAttribute` from dependent body schemas when merging block body schemas in `MergeBlockBodySchemas`
- Add test coverage for the merge behavior

## Motivation

`MergeBlockBodySchemas` copies named `Attributes`, `Blocks`, `TargetableAs`, and other fields from a matched dependent body schema, but previously did not copy `AnyAttribute`.

This matters when a dependent body schema intentionally uses `AnyAttribute` to allow decoding of arbitrary attributes whose names are not known ahead of time. Without propagating `AnyAttribute`, expressions inside those attributes are not decoded and reference origins (e.g. `var.*`) are not collected.

A concrete use case is in `terraform-schema`, where uninstalled remote module blocks need a permissive dependent body schema so that local symbols used as module arguments can still be navigated before `terraform init` downloads child module metadata. That schema relies on `AnyAttribute`, but the merge step was dropping it.

## Test plan

- [x] `go test ./decoder/internal/schemahelper/ -run TestMergeBlockBodySchemas_propagatesAnyAttributeFromDependentBody`
- [x] `go test ./decoder/internal/schemahelper/` (no regressions)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/decoder/internal/schemahelper/block_schema.go
+++ b/decoder/internal/schemahelper/block_schema.go
@@ -31,6 +31,12 @@ func MergeBlockBodySchemas(block *hcl.Block, blockSchema *schema.BlockSchema) (*
 		for name, attr := range depSchema.Attributes {
 			mergedSchema.Attributes[name] = attr
 		}
+		// A dependent body may allow arbitrary attributes (e.g. a module
+		// block whose inputs are not known yet). Propagate AnyAttribute so
+		// that expressions within such attributes are still decoded.
+		if depSchema.AnyAttribute != nil {
+			mergedSchema.AnyAttribute = depSchema.AnyAttribute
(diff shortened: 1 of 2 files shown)


akunaatrium:
I signed the CLA, not sure why it is still pending. Now when I clicked on the signing link, I am getting this:
image

akunaatrium:
Now when I am clicking the link to sign, I see the agreement and the prefilled uneditable fields with my information, giving me the appearance of the agreement being already signed, as I expected. Probably the service which checks whether it is signed or not is defective.

